PR #19, opened 2019-10-24
Remove intermediate artifacts

keefertaylor:
# Background

In XpringKit, there are 3 intermediate artifacts that don't need to be committed. They are:
- XpringKit.xcodeproj: A definition of an XCode Project. This can be generated programmatically by using a tool called `xcodegen` and a `project.yml` file
- Protocol Buffers: These can be regenerated from the `.proto` files provided via submodule
- bundled.js: The is the webpacked JavaScript core, which can be generated via the provided submodule

There are also two build systems:
- Carthage 
- CocoaPods

Carthage is the default that is used for development.

# Changes

- Remove intermediate artifacts and ignore them in `.gitignore`
- Add a new project yml file to generate a cocoapods compatible project, `project_cocoapods.yml`
- Add a new script, `generate_project_cocoapods.sh` which generates a project using the aforementioned yml file
- Add a new script, `clean_project.sh` which removes any intermediate artifacts
- Adds a hook to the CocoaPod's `podspec` file to checkout submodules and run `generate_project_cocoapods.sh` before installing / building the pod
- Changes CI to build / test with Carthage, completely clean, then build / test with cocoapods
- Upgrades XCode version to CI to 11.1 (unrelated to changes, but helped to have local / CI system on same XCode version while debugging)


keefertaylor:
One oddity here is that I could rename:
`generate_project.sh` -> `generate_project_carthage.sh` 
`project.yml` -> `project_carthage.yml`

Which would match `generate_project_cocoapods.sh` and `project_cocoapods.yml`. I kinda like these as they are because it's obvious they are the default we prefer for development, but I can see the argument for renaming them as well.

PR #43, opened 2019-12-05
Documentation Updates for 1.2

keefertaylor:
# Updates documentation for:

## TransactionStatus retrieval feature:
Buckets transactions into 4 states (succeeded, failed, pending, unknown). Unknown is really just an error state that we should never see.

## Blocking Tx Submission
This is a transparent change. Send transactions will now block until they reach a succeeded or failed status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ let xpringClient = XpringClient(grpcURL: remoteURL)
 
 #### Retrieving a Balance
 
-A `XpringClient` can check the balance of an account on the ledger.
+A `XpringClient` can check the balance of an account on the XRP Ledger.
 
 ```swift
 import XpringKit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,9 +153,34 @@ let balance = try! xpringClient.getBalance(for: address)
 print(balance) // Logs a balance in drops of XRP
 ```
 
+### Checking Transaction Status
+
+A `XpringClient` can check the status of an operation on the XRP Ledger.
```
ryangyoung: > the status of an operation

perhaps:

> the status of a transaction

README.md
```diff
@@ -153,9 +153,34 @@ let balance = try! xpringClient.getBalance(for: address)
 print(balance) // Logs a balance in drops of XRP
 ```
 
+### Checking Transaction Status
+
+A `XpringClient` can check the status of an operation on the XRP Ledger. 
+
+Transaction states can be:
```
ryangyoung: I know this is fraught, but maybe we should rephrase this to indicate that these states are _what XpringKit returns_.  I'd be inclined to add a link or two, too, like so (also tweaking the descriptions a bit):

> XpringKit returns the following transaction states:
> - `succeeded`: The transaction was successfully validated and applied to the XRP Ledger.
> - `failed:` The transaction was successfully validated but not applied to the XRP Ledger. Or the operation will never be validated.
>  - `pending`: The transaction has not yet been validated, but may be validated in the future.
>  - `unknown`: The transaction status could not be determined.
> 
> **Note:** For more information, see [Reliable Transaction Submission](https://xrpl.org/reliable-transaction-submission.html) and [Transaction Results](https://xrpl.org/transaction-results.html).


I also changed "operation" to "transaction" throughout to be consistent and because I think the operations are transactions in this case, but feel free to change if that's not correct.

README.md
```diff
@@ -153,9 +153,34 @@ let balance = try! xpringClient.getBalance(for: address)
...
+
+let transactionStatus = xpringClient.getTransactionStatus(for: transactionHash) // TransactionStatus.succeeded
+```
+
 #### Sending XRP
 
-A `XpringClient` can send XRP to other accounts on the ledger.
+A `XpringClient` can send XRP to other accounts on the XRP Ledger.
+
+**Note:** that the payment operation will block until the operation reaches a definitive and irreversible success or failure state.
```
ryangyoung: > the payment operation will block until

Is there a missing word there? What will the operation block?

keefertaylor:
Thank you for detailed comments, @ryangyoung!

PR #45, opened 2019-12-18
Pass through `test` parameter when Decoding X-Addresses 

keefertaylor:
Currently, the test parameter is dropped when decoding X-Addresses. Pass this data through to clients rather than dropping it.

Tested:
- Unit tests

Note that the unit test case is derived from manually checking values on http://xrpaddress.info.

--- 
Underlying core implementation done in: https://github.com/xpring-eng/xpring-common-js/pull/103


keefertaylor:
Note build failure is fixed in  #46

PR #54, opened 2019-12-26
Update to Swift 5

dangell7:
Updated the podspec to swift 5, and changed the `common-js` to my personal fork where I fixed other errors. DO NOT USE my fork, but update your repo with the PR, and then it should work. You should prob also update your submodules to use https: not ssh as some users may not have the ssh enabled and might cause an error.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,8 +1,8 @@
 [submodule "xpring-common-js"]
 	path = xpring-common-js
-	url = git@github.com:xpring-eng/xpring-common-js
+	url = https://github.com/xpring-eng/xpring-common-js.git
 	branch = master
 [submodule "xpring-common-protocol-buffers"]
 	path = xpring-common-protocol-buffers
-	url = git@github.com:xpring-eng/xpring-common-protocol-buffers
+	url = https://github.com/xpring-eng/xpring-common-protocol-buffers
 	branch = master
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "xpring-common-js"]
 	path = xpring-common-js
-	url = https://github.com/dangell7/xpring-common-js.git
+	url = https://github.com/xpring-eng/xpring-common-js.git
 	branch = master
 [submodule "xpring-common-protocol-buffers"]
 	path = xpring-common-protocol-buffers
--- a/XpringKit.podspec
+++ b/XpringKit.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |spec|
 
   spec.swift_versions = [5.0, 4.2]
   spec.requires_arc = true
-  spec.ios.deployment_target = '12.0'
+  spec.ios.deployment_target = '13.2'
   spec.osx.deployment_target = '10.10'
 
   spec.prepare_command = "git submodule update --init --recursive && ./scripts/bundle_js.sh && ./scripts/regenerate_protos.sh"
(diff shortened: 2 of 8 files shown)


dangell7:
SUCCESS!!! Installed on swift 5 repo. I was thinking of adding an example to the sdk, so we can build and run? Any thoughts on that? Or, I can make my project repo public and its just going to be a small wallet function.

.gitmodules
```diff
@@ -1,8 +1,8 @@
 [submodule "xpring-common-js"]
 	path = xpring-common-js
-	url = git@github.com:xpring-eng/xpring-common-js
+	url = https://github.com/dangell7/xpring-common-js.git
```
keefertaylor: can you revert this to point at xpring-eng?  Agree with changing to `https://` though.

XpringKit.podspec
```diff
@@ -6,15 +6,15 @@ Pod::Spec.new do |spec|
   spec.description  = "
     XpringKit provides a Swift based SDK for interacting with Xpring Platform (XRP/ILP)
 
-    XpringKit is part of the Xpring SDK. To learn more, visit: http://github.com/xpring-eng/Xpring-SDK.
+    XpringKit is part of the Xpring SDK. To learn more, visit: http://github.com/dangell7/Xpring-SDK.
```
keefertaylor: ditto

XpringKit.podspec
```diff
@@ -6,15 +6,15 @@ Pod::Spec.new do |spec|
...
   "
 
   spec.homepage     = "http://xpring.io"
   spec.license      = "MIT"
   spec.author       = { "Xpring Engineering" => "xpring@ripple.com" }
-  spec.source       = { :git => "https://github.com/xpring-eng/XpringKit.git", :tag => "1.2.1" }
+  spec.source       = { :git => "https://github.com/dangell7/XpringKit.git", :tag => "1.2.1" }
```
keefertaylor: ditto

keefertaylor:
Glad you were able to get this working. 

Just to be clear, it looks like the stumbling block was that we were using SSH based authentication rather than HTTPS? If so, sorry about that - it's a remnant of getting these libraries building on CI before they were publicly available. 

I have some older demo code here from when we first released the Xpring SDK: https://github.com/xpring-eng/Xpring-SDK-Demo/tree/master/swift/XpringKitDemo. Note this uses Carthage to do dependency management though. You're welcome to update that or push an example there and I'm happy to review / merge those PRs. Also happy to figure out if we can link to any project you may write with this. 

Thanks for providing valuable feedback about the bumps in getting this set up and building. I appreciate that it's been harder than it should have been and thanks for sticking with it. 

From an initial glance at this PR, it looks great. Leaving a few comments and I'll take a deeper look when I properly have some spare time in the next few days.

XpringKit.podspec
dangell7: Done. I'm still learning and this was me trying to get a development pod going. But I think I can just do assign a path in the pod file correct?

.gitmodules
dangell7: Done.

XpringKit.podspec
dangell7: Done.

dangell7:
Also, it was initially the ssh key error, but after that I noticed errors in the common js library. I believe I have a PR there too. Thanks!!

XpringKit.podspec
keefertaylor: > I'm still learning and this was me trying to get a development pod going.

Don't worry, thanks for the valuable feedback and the contribution. 

> But I think I can just do assign a path in the pod file correct?

I'm not sure I follow. You want to fork your own Pod from this one? ?

PR #65, opened 2020-01-07
grpc-Support XpringKit Swift SDK

dangell7:
Long term, xpring-common-protocol-buffers is going to disappear and be replaced with an official and well reviewed implementation by the rippled team.

The implementation in rippled is being done here: ripple/rippled#3159.

To access the new rippled protocol buffers, a new submodule is added. Note that the submodule is cjcobb23/rippled (rather than ripple/rippled) since I can't figure out how to get a submodule to track an open PR. I filed an Asana task to track this update here: https://app.asana.com/0/1143722888069470/1155647690941937.

This change segregates the protocol buffers so that we can do the transition in several phases. As of now:

/generated/*.proto : 'new' protocol buffers from rippled. Unused, as of now.
/generated/legacy/*.proto : The same protocol buffers from xpring-common-protocol-buffers which we have been using.
When we have fully migrated the library to rippled's protocol buffers we can simply delete the legacy folder to delete the migration.


dangell7:
@keefertaylor I added the ripple protocol buffers but it made them in a generated/rpc/v1 folder. Is that correct?

dangell7:
Also, swift doesn't like the duplicate file names so I had to comment out the legacy. How would you suggest I handle that?

scripts/clean_project.sh
```diff
@@ -5,7 +5,7 @@ set -e -o pipefail
 echo "Cleaning XpringKit Project"
 
 rm -rf *.xcodeproj
-rm -rf XpringKit/generated
+rm -rf -r XpringKit/generated/
```
keefertaylor: `-r` is already expressed in `-rf`

dangell7: It was the file name that caused an error. I'll rebuild here in a bit and post the error here.

dangell7: and thanks for the fyi on `-r` / `-rf`

dangell7:
@keefertaylor I wasn't sure how to do a rebase with your prefix-protos with my branch. Idk if I messed it all up. If I did, the only real changes are the regerate.sh file. This was my first `rebase` so I'm still learning.

keefertaylor:
Can you merge master into this branch? 

```
$ git remote add upstream http://github.com/xpring-eng/xpringkit
$ git fetch upstream
$ git merge upstream/master 
$ git push
```

PR #66, opened 2020-01-07
Add `.legacy` to Legacy Protocol Buffers

keefertaylor:
Long term, [xpring-common-protocol-buffers](https://github.com/xpring-eng/xpring-common-protocol-buffers) is going to disappear and be replaced with an official and well reviewed implementation by the rippled team. 

The implementation in rippled is being done here: https://github.com/ripple/rippled/pull/3159.

Swift cannot build if files have the same file name. Therefore, there will be collisions between the rippled protocol buffers and the existing xpring-common-protocol-buffers. This PR adds a `.legacy.{pb, grpc}.swift` postfix to all legacy protocol buffers.


keefertaylor:
CircleCI is having an active outage for OSX clones which is affecting the ability to use homebrew etc which all use git under the covers: https://status.circleci.com/

I'm headed to the gym and dinner and I'll poke this later on when I'm free.

dangell7:
Screen Shot 2020-01-08 at 3 31 50 PM

Its building like this every time I regenerate

PR #67, opened 2020-01-08
[WIP] Skeleton of new `DefaultXpringClient`

dangell7:
Create a new DefaultXpringClient which returns an unimplemented error for all methods. Future PRs will implement these methods piecemeal.

TODO(keefertaylor): Add unit tests.

Background
rippled is going to implement protocol buffer support natively, which is going to use a set of well reviewed protocol buffers. The implementation in rippled is being done here: ripple/rippled#3159.

LegacyDefaultXpringClient: Uses the old protocol buffers
DefaultXpringClient: Uses the new protocol buffers
Once we have a working implementation, the constructor on XpringClient will be changed to take a new boolean parameter, useLegacy, which will control with base implementation is used.


dangell7:
@keefertaylor I know my PR is a mess. I've almost got the new client up and running, but I'm running into an error on the signing of a transaction. I think it's the `signingPublicKey`. In the old version it was a String, i think, and in the new version its Data. I'm wondering if the JS library needs to reflect this change?

dangell7:
@keefertaylor 

```
import JavaScriptCore

/// Allows conversion of `JSValue` objects to `Rpc_V1_Transaction` objects.
internal extension JSValue {
	private enum ResourceNames {
		public static let serializeBinary = "serializeBinary"
	}

	func toSignedTransaction() -> Rpc_V1_Transaction? {
		let javaScriptBytes = self.invokeMethod(ResourceNames.serializeBinary, withArguments: [])!
		guard let bytes = javaScriptBytes.toArray() as? [UInt8] else {
			return nil
		}

		return try? Rpc_V1_Transaction(serializedData: Data(bytes))
	}
}

```

The above is returning nil

dangell7:
See JSVALUE https://developer.apple.com/documentation/javascriptcore/jsvalue/1451666-invokemethod

keefertaylor:
Yeah. This is an annoying internal detail.  We need to land:
https://github.com/xpring-eng/xpring-common-js/pull/125
https://github.com/xpring-eng/xpring-common-js/pull/126

To upgrade the existing code in Xpring-Common-JS. I will try to do that today.

keefertaylor:
#71 has landed which pulls the new JS into this library. 

You should be able to call `signTransaction` which will return a `string`. You're welcome to keep working on this PR or if you'd like, I can make these changes.

dangell7:
@keefertaylor I noticed that the new library returns just the `signatureHex`. Do you want the swift library to create a signed transaction? Or are you leaving that up to the developer?

PR #71, opened 2020-01-13
Upgrade Xpring-Common-JS

keefertaylor:
At head, Xpring-Common-JS provides dual implementations of the `signTransaction` function (`signTransaction` and `signLegacyTransaction`) which supports this library's move to rippled.


keefertaylor:
CC: @dangell7

PR #68, opened 2020-01-08
grpc-Support XpringKit Swift SDK

dangell7:
Long term, xpring-common-protocol-buffers is going to disappear and be replaced with an official and well reviewed implementation by the rippled team.

The implementation in rippled is being done here: ripple/rippled#3159.

To access the new rippled protocol buffers, a new submodule is added. Note that the submodule is cjcobb23/rippled (rather than ripple/rippled) since I can't figure out how to get a submodule to track an open PR. I filed an Asana task to track this update here: https://app.asana.com/0/1143722888069470/1155647690941937.

This change segregates the protocol buffers so that we can do the transition in several phases. As of now:

/generated/.proto : 'new' protocol buffers from rippled. Unused, as of now.
/generated/legacy/.proto : The same protocol buffers from xpring-common-protocol-buffers which we have been using.
When we have fully migrated the library to rippled's protocol buffers we can simply delete the legacy folder to delete the migration.

PR #72, opened 2020-01-14
Migrate existing code to `Legacy` subfolders

keefertaylor:

I propose doing this upgrade in a few stages:
- [x] Migrate existing code to a `legacy` subdirectory and namespace (This PR)

----

Analog in Xpring-JS: https://github.com/xpring-eng/XpringKit/pull/68


keefertaylor:
CC: @dangell7

XpringKit/Legacy/Io_Xpring_XRPLedgerAPIServiceClient+LegacyNetworkClient.swift
```diff
@@ -0,0 +1,3 @@
+/// Conforms XRPLedgerServiceClient to NetworkClient protocol.
+extension Io_Xpring_XRPLedgerAPIServiceClient: LegacyNetworkClient {
+}
```
hbergren: I'm unfamiliar with Swift. What does this `extension` do?

keefertaylor: Swift extensions let you add additional functionality to classes. This is useful (and sometimes scary).

What we're doing here is taking a generated class (`Io_Xpring_XRPLedgerAPIServiceClient` and conforming it to a protocol). Protocols are an analog to an `interface` in java or javascript. The protocol happens to implement these methods already, so this extension is empty.

You generally can use extensions to add helper methods, or you can use them to conform types which you don't own (system classes, generated code, other libraries) to protocols. There are more less-sane ways to use them too. 

See more: https://docs.swift.org/swift-book/LanguageGuide/Extensions.html

PR #75, opened 2020-01-16
GRPC V1 Move Client to Legacy

dangell7:

Moved the rest of the clients and tests to legacy after receiving errors.

LegacyFakeClient
LegacyDefaultClient
LegacyIntegration
LegacyReliableSubmission
LegacyClient
LegacyClientDecorator

### Context of Change

Refactored out legacy Io_Xpring functions/classes

### Type of Change

- [X ] Refactor (non-breaking change that only restructures code)

## Before / After

I moved all the current clients, decorators and tests to legacy because I received errors when running tests and functions/classes using the new Rpc_V1. Also, I would think that if we are going to deprecate the functions/classes in the future, that the current implementations should be removed completely.

## Test Plan

All tests passed.

## Future Tasks

Confirm Test... Deprecate old Io_Xpring functions/classes


keefertaylor:
I apologize that we ended up duplicating work on this. 

I think if you merge in `master` this PR will become a no-op, since I landed: https://github.com/xpring-eng/XpringKit/pull/72

PR #77, opened 2020-01-16
add bytesToArray to Utils

dangell7:

Add Bytes to Array Function to Utils for converting Hex String to bytes array

### Context of Change

Architecture was influenced by @keefertaylor 

### Type of Change

- [ X] New feature (non-breaking change which adds functionality)

## Before / After

Not Sure. Use whenever converting string hex to data

## Test Plan

All Tests Passed

## Future Tasks

Add to getTx

### DIFF
--- a/Tests/UtilsTest.swift
+++ b/Tests/UtilsTest.swift
@@ -3,242 +3,269 @@ import XpringKit
 
 class UtilsTest: XCTestCase {
 
-  // MARK: - isValid
+    // MARK: - isValid
 
-	func testIsValidAddressValidClassicAddress() {
-    XCTAssertTrue(Utils.isValid(address: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1"))
-	}
-
-  func testIsValidAddressValidXAddress() {
-    XCTAssertTrue(Utils.isValid(address: "XVLhHMPHU98es4dbozjVtdWzVrDjtV18pX8yuPT7y4xaEHi"))
-  }
-
-	func testIsValidAddressInvalidAlphabet() {
-    XCTAssertFalse(Utils.isValid(address: "1EAG1MwmzkG6gRZcYqcRMfC17eMt8TDTit"))
-	}
-
-	func testIsValidAddressInvalidChecksumClassicAddress() {
-    XCTAssertFalse(Utils.isValid(address: "rU6K7V3Po4sBBBBBaU29sesqs2qTQJWDw1"))
-	}
-
-  func testIsvValidAddressInvalidChecksumXAddress() {
-    XCTAssertFalse(Utils.isValid(address: "XVLhHMPHU98es4dbozjVtdWzVrDjtV18pX8yuPT7y4xaEHI"))
-  }
-
-	func testIsValidAddressInvalidCharacters() {
-    XCTAssertFalse(Utils.isValid(address: "rU6K7V3Po4sBBBBBaU@#$%qs2qTQJWDw1"))
-	}
-
-	func testIsValidAddressTooLong() {
-    XCTAssertFalse(Utils.isValid(address: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1"))
-	}
-
-	func testIsValidAddressTooShort() {
-    XCTAssertFalse(Utils.isValid(address: "rU6K7V3Po4s2qTQJWDw1"))
-	}
-
-  // MARK: - isValidXAddress
-
-  func testIsValidXAddressWithXAddress() {
-    // GIVEN a valid X-Address.
-    let address = "XVfC9CTCJh6GN2x8bnrw3LtdbqiVCUvtU3HnooQDgBnUpQT"
-
-    // WHEN the address is validated for being an X-Address.
-    let isValid = Utils.isValidXAddress(address: address)
-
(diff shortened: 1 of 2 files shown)


XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
-  }
+    /// Convert the given transaction blob to a transaction hash.
+    ///
+    /// - Parameter transactionBlobHex: A hexadecimal encoded transaction blob.
+    /// - Returns: A hex encoded hash if the input was valid, otherwise undefined.
+    public static func toTransactionHash(transactionBlobHex: Hex) -> Hex? {
+        return javaScriptUtils.toTransactionHash(transactionBlobHex: transactionBlobHex)
+    }
+
+    /// Convert the given string to a bytes array UInt8.
```
keefertaylor: "to a Uint8 byte array."

XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
+    ///
+    /// - Parameter hex: A string.
+    /// - Returns: A byte array in [UInt8].
+    public static func toByteArray(hex: String ) -> [UInt8] {
```
keefertaylor: There's a `Hex` type alias you can use instead of `String`

keefertaylor: Nit: remove  space before `)`

XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
+        // remove "-" from Hexadecimal
+        let hexString = hex.removeWord( "-" )
```
keefertaylor: Remove spaces after `(` and before `)`. Here and elsewhere in this method. This is consistent with style throughout the existing code.

keefertaylor: In practice, do we care about '-'. I don't think there's every any input into this functions that are dashes and this feels like an odd one-off check to me.

XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
+        let size = hexString.count / 2
+        var result: [UInt8] = [UInt8]( repeating: 0, count: size ) // array with length = size
+        for i in stride( from: 0, to: hexString.count, by: 2 ) {
+            let subHexStr = hexString.subString( i, length: 2 )
+            guard let _uint8 = UInt8( subHexStr, radix: 16 ) else {
```
keefertaylor: why prefix this with `_`

XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
+                return []
+            }
+            result[ i / 2 ] = _uint8
+        }
+        return result
+    }
+}
+extension String {
+
+    func subString( _ from: Int, length: Int ) -> String {
```
keefertaylor: I'd prefer we just use `prefix` to get the first two elements, rather than re-implementing a `substring` method:
https://developer.apple.com/documentation/swift/string/2894830-prefix

This method is essentially implementing [String's subscript](https://developer.apple.com/documentation/swift/string/2943910-subscript) method. Would prefer we use the official definitions provided by foundation rather than hand rolling our own.

XpringKit/Utils.swift
```diff
@@ -1,66 +1,132 @@
...
+        var result = ""
+        for ( idx, char ) in self.enumerated() {
+            if idx >= from && idx < to {
+                result.append( char )
+            }
+        }
+        return result
+    }
+
+    func removeWord( _ word: String ) -> String {
```
keefertaylor: It's a bit opaque what this method is doing. I think that you could just use [dropFirst](https://developer.apple.com/documentation/swift/string/2893343-dropfirst).

Tests/UtilsTest.swift
```diff
@@ -3,242 +3,269 @@ import XpringKit
...
+        XCTAssertEqual(
+            hexBytesArray.toHex(),
+            "9A88C8548E03958FD97AF44AE5A8668896D195A70CF3FF3CB8E57096AA717135"
+        )
+    }
 
-    // THEN the hash is nil.
-    XCTAssertNil(transactionHash)
-  }
+    func testToBytesArrayInvalidTransactionHash() {
```
keefertaylor: This test feels odd to me. If we give an invalid input, I'd expect either:
- `nil` output
- an exception to be thrown for an invalid input 

I'd probably prefer to throw an exception rather than making the output optional but happy to hear your thoughts if you have a strong opinion.

I would expect 
```
> Utils.toByteArray(hex: "")
[]
```

There's no way to distinguish this case from the empty input case with the current implementation.

Tests/UtilsTest.swift
```diff
@@ -3,242 +3,269 @@ import XpringKit
...
-    let transactionHash = Utils.toTransactionHash(transactionBlobHex: transactionBlobHex)
+    // MARK: - toBytesArray
 
-    // THEN the transaction blob is as expected.
-    XCTAssertEqual(
-      transactionHash,
-      "7B9F6E019C2A79857427B4EF968D77D683AC84F5A880830955D7BDF47F120667"
-    )
-  }
+    func testToBytesArrayValidTransactionHash() {
```
keefertaylor: Please add a test case for the empty string as well.

XpringKit/Utils.swift
dangell7: I couldn't get the prefix to work, It seems its a custom function. You're welcome to refactor the function.

PR #81, opened 2020-01-23
Implement a signer for rippled protocol buffers

keefertaylor:

Implement a Signer which uses the rippled protocol buffers and xpring-common-js. 

### Context of Change

#### Status
[x] [Generate new versions of the protocol buffers](https://github.com/xpring-eng/XpringKit/pull/68)
   - [x] [Create a new `NetworkClient` with fakes](https://github.com/xpring-eng/XpringKit/pull/86)
   - [x] [Create a new `DefaultXpringClient` with tests](https://github.com/xpring-eng/XpringKit/pull/87)
   - [ ] Read only calls for new `DefaultXpringClient` (WIP in this PR)

### Type of Change

## Before / After

N/A - Code isn't yet hooked up.

## Test Plan
 New unit tests pass on CI.


XpringKit/JavaScript/JavaScriptSerializer.swift
```diff
@@ -31,11 +36,11 @@ internal class JavaScriptSerializer {
     return walletClass.construct(withArguments: [ wallet.publicKey, wallet.privateKey])
   }
 
-  /// Serialize a transaction to a JavaScript object.
+  /// Serialize a legacy transaction to a JavaScript object.
```
hbergren: Should this have legacy in the comment, since this serialize takes an `Rpc_V1_Transaction`?

keefertaylor: Done. Botched copy / paste job on my part.

PR #83, opened 2020-01-24
Update xpring-common-js for gRPC-Web

Stormtv:
This add compatibility to the upcoming changes to the webpack output of xpring-common-js.

### Context of Change
bundled was renamed to index inside of xpring-common-js.

This was done in order to have the typescript definitions to match up with the outputted file. There was no easy way to rename the index typescript definition within the tsconfig. Although we could have a script that renames it after the webpack process is completed.

In addition the global var EntryPoint was changed to `XpringCommonJS` this allows people who are consuming the webpack distribution to have a meaningful variable name.

### Type of Change
- [X] Refactor (non-breaking change that only restructures code)

## Before / After
Now points to proper var / resource file

## Test Plan
CI Passes


XpringKit/JavaScript/XRPJavaScriptLoader.swift
```diff
@@ -61,7 +61,7 @@ internal enum XRPJavaScriptLoader {
     // TODO(keefer): Drop context parameter when context becomes a singleton.
     public static func load(_ resourceName: String, from context: JSContext) -> JSValue {
         guard
-            let entrypoint = context.objectForKeyedSubscript("EntryPoint"),
+            let entrypoint = context.objectForKeyedSubscript("XpringCommonJS"),
```
keefertaylor: Please update comment on line 53.

keefertaylor: Should we call this variable something besides entrypoint? `xpringCommon`, `xpringCommonJS` or `jsRoot`? I'm not sure what the right terminology would be.

Stormtv:
@keefertaylor Hopefully these name changes made it clearer.

scripts/clean_project.sh
```diff
@@ -7,5 +7,6 @@ echo "Cleaning XpringKit Project"
 rm -rf *.xcodeproj
 rm -rf XpringKit/generated
 rm -f XpringKit/Resources/bundled.js
```
keefertaylor: You can remove line 9. 

You could also just genericize this and make it  `rm-rf XpringKit/Resources` and not have to think about this again.

Stormtv: Smart, made that change :)

Stormtv:
> Looks reasonable to me. I assume you've tested and it works?

It passed all test on the latest version of xCode (CI takes forever but should pass soon)

keefertaylor:
Thank you!

PR #86, opened 2020-01-27
Network Clients on rippled protocol buffers

keefertaylor:
Create a new `NetworkClient` interface, conform generated gRPC client to `NetworkClient` and add a  `FakeNetworkClient` for testing. 

These classes are just light syntactical wrappers around the underlying gRPC implementation which give us: 
- A facade abstraction, if we ever wanted to move off of gRPC
- Promise based callbacks on gRPC
- Testability through a Fake

# See also:
Analog in JavaScript: https://github.com/xpring-eng/Xpring-JS/pull/111.

# Background

# Status
   - [x] Create a new `NetworkClient` with fakes (this PR)
   - [ ] Create a new `DefaultXpringClient` with tests
   - [ ] Read only calls for new `DefaultXpringClient`


XpringKit/NetworkClient.swift
```diff
@@ -0,0 +1,30 @@
...
+  /// - Throws: An error if there was a problem communicating with the XRP Ledger.
+  /// - Returns: A response from the ledger acknowledging the transaction submission.
+  func submitTransaction(_ request: Rpc_V1_SubmitTransactionRequest) throws -> Rpc_V1_SubmitTransactionResponse
+
+  /// Retrieve transaction status for a given transaction hash from the XRP Ledger.
+  ///
+  /// - Parameter request: Request parameters for retrieving transaction status.
+  /// - Throws: An error if there was a problem communicating with the XRP Ledger.
+  /// - Returns: A `TransactionStatus`  for the transactioon.
+  func getTx(_ request: Rpc_V1_GetTxRequest) throws -> Rpc_V1_GetTxResponse
```
hbergren: :+1: Good job documenting this so well with in-line comments!

PR #92, opened 2020-01-28
Add conversion function for Hex to Bytes

keefertaylor:

Add a conversion function for a hexadecimal string to bytes. 

### Context of Change

rippled is going to implement protocol buffer support natively, which is going to use a set of well reviewed protocol buffers. The implementation in rippled is being done here: ripple/rippled#3159.

rippled requires us to use protocol buffers that contain raw bytes. We provide APIs that use hex strings. This functionality allows us to interoperate.

### Type of Change

## Before / After
N/A

## Test Plan

CI. 

Added tests for cases:
- Upper / Lower Case Hexadecimal strings
- Invalid hex string lengths (because 1 byte = 2 hex characters, string length must be a multiple of 2)
- Invalid hex characters


XpringKit/Hex+ByteArray.swift
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    var startIndex = self.startIndex
+    return stride(from: 0, to: count, by: 2).compactMap { _ in
+      let endIndex = index(startIndex, offsetBy: 2, limitedBy: self.endIndex) ?? self.endIndex
+      defer { startIndex = endIndex }
+
+      return UInt8(self[startIndex..<endIndex], radix: 16)
+    }
+  }
```
hbergren: This `stride` and `compactMap` is a little complicated. Does it deserve a code comment?

This reads to me as "Loop through this thing by steps of two, and then... do some stuff? Not sure about the specifics.

keefertaylor: Updated. This code was mostly just refactored from a stack overflow post so it was a bit hard. 

It now does the following (note that we're implementing a new method on `Hex`, which is an alias to the builtin `String type):
- strides through the characters from 0...count, by two
- feeds that index to a compactMap function (this lets us safely return nil values, because line 19 returns an optional). We know this won't happen because of the guard on line 10. 
- Builds `string.Index` characters from the index fed in by stride. Basically, strings in swift have a complex index type so they can handle different encodings
- maps those two hex characters to a byte and returns it from the map function.

keefertaylor: I also added comments to make it more obvious what is happening.

PR #93, opened 2020-01-28
Implement getRawTransactionStatus

keefertaylor:

Implement `getRawTransactionStatus` with the new rippled protocol buffers. 

Analog in JS: https://github.com/xpring-eng/Xpring-JS/pull/131

### Context of Change

rippled implemented protocol buffer support natively, which uses a new and incompatible set of protocol buffers. The implementation in rippled was completed in: https://github.com/ripple/rippled/pull/3159.

`XpringClient` uses a decorator pattern where each implementation must conform to an interface. The interface required for the `getRawTransactionStatus` method requires it return a `RawTransactionStatus`. 

This PR implements the `getRawTransactionStatus` method of the decorator. 

*Status*

   - [x] [Create a new `NetworkClient` with fakes](https://github.com/xpring-eng/XpringKit/pull/86)
   - [x] [Create a new `DefaultXpringClient` with tests](https://github.com/xpring-eng/XpringKit/pull/87)
   - [ ] Read only calls for new `DefaultXpringClient` (WIP in this PR)
- [x] [Add a boolean parameter to switch between implementations](https://github.com/xpring-eng/XpringKit/pull/90)

### Type of Change

## Before / After

No change. This new class is not exposed to the world or hooked up yet. 

## Test Plan

CI Continues to work.


keefertaylor:
Sorry for premature review request. There's a dependency before this can land.

XpringKit/DefaultXpringClient.swift
```diff
@@ -1,3 +1,5 @@
+import Foundation
```
hbergren: What is `Foundation`?

If this was JS code, I would say this looks like an unused import, but idk how this works in Swift-land.

XpringKit/DefaultXpringClient.swift
```diff
@@ -78,6 +80,15 @@ extension DefaultXpringClient: XpringClientDecorator {
   /// - Throws: An error if there was a problem communicating with the XRP Ledger.
   /// - Returns: The status of the given transaction.
   public func getRawTransactionStatus(for transactionHash: TransactionHash) throws -> RawTransactionStatus {
-    throw XRPLedgerError.unimplemented
+    let transactionHashBytes = try transactionHash.toBytes()
```
hbergren: Does the `try` here imply it can fail? What gets stored in `transactionHashBytes` on failure?

Or does that throw an error, and we stop executing (hence the `throws` in the method signature?)

XpringKit/DefaultXpringClient.swift
keefertaylor: All Swift programs import the [Swift Standard Library](https://developer.apple.com/documentation/swift/swift_standard_library/) by default. This lets us use things like numbers or Strings. We haven't needed to import anything before now because we're not using anything more advanced. 

[Foundation](https://developer.apple.com/documentation/foundation) is a core framework Apple provides. In it, there is a `Data` type, which is an opaque wrapper for a collection of bytes. A data object is made on line 84  which takes the array of bytes (typed as '[Uint8]` in swift) returned from `toHex()` to a `Data`.
---
Since you piqued my curiousity, I went out and found out to my surprise that [class member imports are possible](https://nshipster.com/import/) even though I've *never* seen Swift code that uses them.  I googled 'Swift style guide' and of the top 3 hits, [LinkedIn's guide](https://github.com/linkedin/swift-style-guide) makes no mention of imports, whereas Ray Wenderlich's (a well known tutorial site in the swift community) [demonstrates whole module imports](https://github.com/raywenderlich/swift-style-guide#minimal-imports) but does not explicitly recommend them. [Google's guide](https://google.github.io/swift/#import-statements) is explicit about this, saying:

> Imports of whole modules are preferred to imports of individual declarations or submodules.
There are a number of reasons to avoid importing individual members:
> - There is no automated tooling to resolve/organize imports.
> - Existing automated tooling (such as Xcode’s migrator) are less likely to work well on code that imports individual members because they are considered corner cases.
> - The prevailing style in Swift (based on official examples and community code) is to import entire modules.

hbergren: Great explanation. I feel like I understand Swift a little better now!

XpringKit/DefaultXpringClient.swift
keefertaylor: On line 81, I've added a `throws` keyword, which tells callers of the function they need to handle an error which may be thrown. 

In this case `transactionHashBytes` is typed as a byte array (`[UInt8]`) rather than an optional byte array (`[UInt8]?`). 

There are three ways to handle errors in Swift:
- `try`: Handle a thrown exception (either by propagating upwards with your method signature, or with a  `catch` block.
- `try?`:  Ignore any thrown exceptions and return `nil` (Swift's version of null) if an exception is thrown. Note that this would cause the type compiler to infer `transactionHashBytes` as an optional byte array, which would change our logic below.
- `try!`: Crash the program if an exception is thrown. This is poor form to use in anywhere that isn't test code.

Swift enforces exception handling at compile time for safety. So if I removed `try` from this line, the program would fail to compile because I haven't handled a possible exception. Likewise, if I removed `throws` from line 82, I would also get a compile time error. This in turn means that anyone who calls this method will have to use `try` and handle exceptions appropriately.

keefertaylor:
Thanks for the Swift questions!

PR #98, opened 2020-01-28
Integration tests for read gRPC calls on rippled

keefertaylor:

Implement integration tests for the read only methods of rippled. 

Analog in JS: https://github.com/xpring-eng/Xpring-JS/pull/140

### Context of Change

rippled implemented protocol buffer support natively, which uses a new and incompatible set of protocol buffers. The implementation in rippled was completed in: https://github.com/ripple/rippled/pull/3159.

I implemented functionality for `getBalance` and `getTransactionStatus`. These new tests wire these against a live server. In practice, these tests are not hermetic, but they have proven not to be too flakey thus far. 

The server on the other side is a rippled node that I have stood up on an Ubuntu AWS box. It is publicly accessible and attached to the testnet network. 

These are analogs to the legacy tests in the same file. 

### Type of Change

[x] Tests

## Before / After

Code had no integration tests. Now code is tested! 

## Test Plan

I will not be testing the tests.


keefertaylor:
(Grabbing another reviewer since @tedkalaw is at Jury duty)

Tests/IntegrationTests.swift
```diff
@@ -35,17 +41,36 @@ final class IntegrationTests: XCTestCase {
...
+
+  func testGetBalance_legacy() {
+    do {
+      _ = try legacyClient.getBalance(for: Wallet.testWallet.address)
+    } catch {
+      XCTFail("Failed retrieving balance with error: \(error)")
+    }
+  }
+
+  func testSendXRP_legacy() {
```
stephengu: is there a send for the new client?

keefertaylor: Implemented in PR #99, which has these commits in it.

PR #100, opened 2020-01-28
Migrate away from BigInt

keefertaylor:

UInt64 allows appropriate levels of precision for all amounts of XRP. Remove BigInt framework dependency.

### Context of Change

There are 10e19 drops of XRP in the world, which can be safely represented with a 64 bit integer. rippled's protos also use uint64s to represent drop amounts. 

To be more consistent, migrate XpringKit to use the build in UInt64 type, rather than a 3rd party framework that provides arbitrary precision integers. 

This is a breaking change as it changes a public API, but the benefits outweigh the costs from a security / maintainability perspective. There's also no great way to change a BigUInt to a UInt64, so compatibility with rippled's protocol buffers will also be hard if this change isn't made. 

### Type of Change

## Before / After
Users will have to upgrade their code to use UInt64s rather than BigUInts

## Test Plan

Unit tests are updated to new types and continue to pass.


XpringKit/Legacy/LegacyDefaultXpringClient.swift
```diff
@@ -52,13 +50,17 @@ extension LegacyDefaultXpringClient: XpringClientDecorator {
...
-  public func getBalance(for address: Address) throws -> BigUInt {
+  public func getBalance(for address: Address) throws -> UInt64 {
     guard Utils.isValidXAddress(address: address) else {
       throw XRPLedgerError.invalidInputs("Please use the X-Address format. See: https://xrpaddress.info/.")
     }
 
     let accountInfo = try getAccountInfo(for: address)
-    return BigUInt(stringLiteral: accountInfo.balance.drops)
+    guard let drops = UInt64(accountInfo.balance.drops) else {
+      throw XRPLedgerError.malformedResponse("Couldn't parse drops to a balance")
```
hbergren: 👍 Nice thinking about edge cases / errors

keefertaylor: Fun fact, swift uses Optional types. So the type returned from `UInt64(accountInfo.balance.drops)` is actually a Uint64 or nil (in Swift, we'd write this type as `UInt64?`). 

This function defines itself to return a `UInt64` rather than than an optional UInt64 (`UInt64?`) on line 53 so we can't not handle this edge case (it'll be a compiler error). The `guard let` syntax is called optional binding which checks that an optional has a `some` value instead of a `nothing` value.

hbergren: This is very similar to Elm, a "compile-to-JS" language based on FRP and Haskell. Very neat.

PR #102, opened 2020-02-04
[1.3.0] Cut Release

keefertaylor:
Cut a release for 1.3.0. 

Add CHANGELOG.md to keep track of versions going forward.


CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## 1.3.0 - Feb 4, 2020
```
hbergren: 👍 on adding a changelog

XpringKit/JavaScript/JavaScriptSigner.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Returns: A `SignedTransaction` derived from the inputs.
+    public func sign(_ transaction: Rpc_V1_Transaction, with wallet: Wallet) -> [UInt8]? {
+        guard let javaScriptTransaction = javaScriptSerializer.serialize(transaction: transaction) else {
+            return nil
+        }
+        let javaScriptWallet = javaScriptSerializer.serialize(wallet: wallet)
+
+        let javaScriptSignedBytes = signTransactionFunction.call(withArguments: [javaScriptTransaction, javaScriptWallet])!
+        let signedBytes = javaScriptSignedBytes.toArray()
+        return signedBytes as? [UInt8]
```
hbergren: What's the `as?` here?

keefertaylor: This needed master merged into it so this is already committed. 

To answer your question:
`as` is a cast operator. You can have:
- `as`: perform a cast that we know will always succeed due to static types (ie. `0 as Integer`)
- `as?`: attempt a cast and if the cast fails, assign nil:
```swift
let x = 0 as? Integer // x = 0, type Integer? (Optional integer)
let y = 0.5 as? Integer // y = 0, type Integer? (Optional integer)
```
- `as!`: force cast, and crash if things don't go great. This is generally poor form to use in code:
```swift
let x = 0 as! Integer // x = 0, type Integer 
let y = 0.5 as? Integer // program will crash with runtime exception. 
```

hbergren: Neat. So this works differently than `as` in TypeScript, because Swift has real types.

In TypeScript, the `as` operator tells the compiler to treat the value as that type, but it doesn't really perform a cast or do anything else.

PR #104, opened 2020-02-06
Propagate additional failure status codes to `LegacyDefaultXpringClientTest`

keefertaylor:

@stephengu asked for multiple test cases on failure codes in XpringKit's `DefautXpringClientTests`. This PR propagates the same logic to the legacy tests.  

Resolve TODO for this. The tests were added to:
JS: https://github.com/xpring-eng/Xpring-JS/pull/175
Java: https://github.com/xpring-eng/xpring4j/pull/90

### Context of Change

The rippled server returns [many codes](https://xrpl.org/transaction-results.html) which mean 'failure'. These are all prefixed in different ways. This PR tests all failure prefixes. 

This change was asked for in: https://github.com/xpring-eng/XpringKit/pull/94#pullrequestreview-351145809

### Type of Change

## Before / After

No change

## Test Plan

N/A

### DIFF
--- a/XpringKit.xcodeproj/project.pbxproj
+++ b/XpringKit.xcodeproj/project.pbxproj
@@ -701,6 +701,8 @@
 			isa = PBXProject;
 			attributes = {
 				LastUpgradeCheck = 1020;
+				TargetAttributes = {
+				};
 			};
 			buildConfigurationList = C0A52FC1F46AB2CDA8428A29 /* Build configuration list for PBXProject "XpringKit" */;
 			compatibilityVersion = "Xcode 10.0";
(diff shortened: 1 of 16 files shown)


XpringKit.xcodeproj/project.pbxproj
```diff
@@ -701,6 +701,8 @@
 			isa = PBXProject;
 			attributes = {
 				LastUpgradeCheck = 1020;
+				TargetAttributes = {
```
keefertaylor: This is a generated file, I think that either my local version of XCode / the generation tool were updated. This should be a no-op.

PR #110, opened 2020-02-10
Add `RippledFlags` class

keefertaylor:

XpringKit will need to interact with flags to work with rippled. Examples of these interactions include:
- Checking for partial payments
- Requiring canonical signatures

### Context of Change

Swift provides an [OptionsSet](https://developer.apple.com/documentation/swift/optionset) which makes working with bit sets easy. 

This is a trio of changes implemented across Xpring SDK:
- JS: https://github.com/xpring-eng/Xpring-JS/pull/181
- Java: https://github.com/xpring-eng/xpring4j/pull/94
- Swift: https://github.com/xpring-eng/XpringKit/pull/110

### Type of Change

## Before / After

N/A - This new struct has no clients.

## Test Plan

N/A


hbergren:
Should there be any code changes in this PR or just `pbxproj` file changes?

keefertaylor:
@hans yes, sorry I did not add those files. 

PTAL.

keefertaylor:
(- Hans who is out, + a new devx reviewer)

XpringKit/RippledFlags.swift
```diff
@@ -0,0 +1,15 @@
+/// Flags used in rippled transactions.
+///
+/// These are only flags which are utilized in Xpring SDK. For a complete list of flags, see:
+/// https://xrpl.org/transaction-common-fields.html#flags-field.
+public struct RippledFlags: OptionSet {
+  public let rawValue: UInt32
+
+  /// Flag values.
+  public static let tfPartialPayment = RippledFlags(rawValue: 1 << 17)
```
hbergren: Is this just for detecting partial payments? SDK does not support sending partial payments, correct?

I'm actually curious why we _need_ to detect partial payments. Shouldn't looking at the `amount` field of the transaction be enough to determine how much we actually received?

keefertaylor: > Is this just for detecting partial payments? 
Correct

> SDK does not support sending partial payments, correct?
Correct

> I'm actually curious why we need to detect partial payments. 
We're trying to bucket transactions because we want to make the payments use case dead simple. 

> Shouldn't looking at the amount field of the transaction be enough to determine how much we actually received?

I guess we could do that. That seemed slightly more complicated than just checking for this flag and the marginal value felt pretty low.

Aside, and recognizing that this is a weak justification:  We need flags anyway (we should really be setting [tfCanonicalSig](https://xrpl.org/transaction-common-fields.html#global-flags) for instance so this seemed like an okay way to do this.

keefertaylor: BTW: This is how this class ultimately gets used: 
https://github.com/xpring-eng/Xpring-JS/pull/193

hbergren: So what happens if someone sends you a partial payment, and we're using `xpring-sdk`? We receive the transaction, say we can't parse it, but then when you fetch your account balance again it'll show it as incremented (from that transaction?)

keefertaylor: This API is for `getTransactionStatus`, so you give us a transaction hash and we give you an enum. It's maybe better to rename this method to `getPaymentStatus`. 

`getBalance` will just go straight to the ledger. 

There's no incoming hooks for this stuff; so you would need to BYOTxHash either by pulling it from TxHistory or by some other means.

keefertaylor:
Followed up with @hbergren  offline and we're happy to merge this PR.

PR #117, opened 2020-02-28
Upgrade for breaking changes in rippled protocol buffers

keefertaylor:

Update to be compatible with breaking protocol buffer changes in rippled.  Breaking changes were introduced in  https://github.com/ripple/rippled/pull/3254. .

### Context of Change

No logic changes are introduced. This change does not touch the public API. 

Changes for protocol buffer compatibility and upgrade xpring-common-js to a version that has compatibility with the new protocol buffers

Changes in the protocol buffers are:
- Change folder / package location of files
- Introduce new layers of abstraction internally.

### Type of Change

## Before / After

Compatible with latest rippled. 

## Test Plan
 CI


XpringKit/ReliableSubmissionXpringClient.swift
```diff
@@ -48,7 +48,6 @@ extension ReliableSubmissionXpringClient: XpringClientDecorator {
       Thread.sleep(forTimeInterval: ledgerCloseTime)
 
       latestLedgerSequence = try getLatestValidatedLedgerSequence()
-      print("got latest: \(latestLedgerSequence)")
```
hbergren: Not a useful print?

keefertaylor: Yeah, this is just left over and I happened to notice and wanted to clean up.

keefertaylor:
> LGTM, we are ignoring generated files in xpring-js is there a reason we don't ignore them in XpringKit?

Carthage is a popular swift dependency manager that checks out code from github and builds it on the CLI. Users were complaining that they were getting errors because they didn't have appropriate tooling installed (protocol buffer compiler, grpc plugin, etc) so we check in the generated artifacts.

PR #121, opened 2020-02-28
Update documentation to point towards hosted nodes rather than shimming service

keefertaylor:
We're confident in the rippled node's ability to serve the demo SDK and want to point SLOHack users to these endpoints.


keefertaylor:
CC: @dpurdy

PR #116, opened 2020-02-27
Adds accountExists method to XpringClient

amiecorso:
## High Level Overview of Change
Implements accountExists method in XpringClient class.  Third and final implementation, such that this method now exists across all three current SDK flavors.

### Context of Change
accountExists was implemented directly in the xpring-wallet-system project, but should belong to the XpringClient class.  

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)

## Before / After
accountExists now exists in XpringClient (across all three SDKs)

## Test Plan
CI

PR #124, opened 2020-03-03
(2/2) Implement skeleton for getTransactionHistory

keefertaylor:

Implement a skeleton of `getTransactionHistory` and tests inside of XpringClient.

JS: https://github.com/xpring-eng/Xpring-JS/pull/212
Java: https://github.com/xpring-eng/xpring4j/pull/116
Swift: https://github.com/xpring-eng/XpringKit/pull/124

### Context of Change

rippled recently implemented this RPC which we'd like to use.

Note that this functionality is not exposed in `XpringClient`, so this is still internal only. Currently `Transaction` objects are just empty containers. 

A future PR will:
- Update `Transaction` to have fields and to set them correctly from the protocol buffers
- Expose this as a public API

### Type of Change

## Before / After

N/A - still internal only functionality. 

## Test Plan

New unit tests provided for CI.


Tests/DefaultXpringClientTest.swift
```diff
@@ -418,6 +418,66 @@ final class DefaultXpringClientTest: XCTestCase {
     }
   }
 
+  // MARK: - TransactionHistory
```
hbergren: What does this `MARK` mean?

Tests/DefaultXpringClientTest.swift
```diff
@@ -418,6 +418,66 @@ final class DefaultXpringClientTest: XCTestCase {
...
+  func testTransactionHistoryWithSuccess() {
+    // GIVEN a Xpring client which will successfully return transactionHistory mocked network call.
```
hbergren:
```suggestion
    // GIVEN a Xpring client which will successfully return a transactionHistory mocked network call.
```

Tests/Helpers/Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse+Test.swift
```diff
@@ -1,9 +1,22 @@
 import XpringKit
 
+// TODO(keefertaylor): Refactor Org_Xrpl_Rpc_V1_AccountAddress to be a static test helper object to avoid repetition.
 extension Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse {
   static let testTransactionHistoryResponse = Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse.with {
     $0.account = Org_Xrpl_Rpc_V1_AccountAddress.with {
       $0.address = .testAddress
     }
+
+    $0.transactions = [
```
hbergren: What does the `$0` here mean?

amiecorso: yeah I'd love to know this too

Tests/DefaultXpringClientTest.swift
amiecorso: keefer explains this here: https://github.com/xpring-eng/XpringKit/pull/116#discussion_r387110813

Tests/Helpers/Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse+Test.swift
keefertaylor: This is kinda complicated. 

this method is defined as:
```
    public static func with(_ populator: (inout Self) throws -> ()) rethrows -> Self
````

What this says is: 
- `with` is a static method  (`public static func with(`)
- which takes a closure (you could also call this an anonymous function or lambda in other languages) (`_ populator: (inout Self) throws -> ())
- This block takes an inout parameter of type self
- And the whole function returns an instance of `Self` (`-> Self`)

So, we could write something like this (long form):
```
Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse.with((populator: Self) in 
  populator.transactions = <...>
)
````

Having the last parameter of a method be a closure is pretty common in swift, so we actually have special syntax when we write outside of the method ("trailing closure syntax"):

```
Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse.with() { (populator: Self) in 
  populator.transactions = <...>
}
```

Trailing closure syntax also lets us drop the `()` for brevity if there are no other parameters. And we can drop the type annotation in the parameter to the closure:
```
Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse.with { populator in 
  populator.transactions = <...>
}
```

Swift also lets us drop the names of parameters and refer to them as `$<param number` in closures, so we can take this one step further:

```
Org_Xrpl_Rpc_V1_GetAccountTransactionHistoryResponse.with { 
  $0.transactions = <...>
}
```

(This is in fact the recommended syntax from the Swift guide: https://github.com/apple/swift-protobuf#quick-start)

keefertaylor: In general, Swift has some gnarly closure syntax if you don't know what you're looking at. The definitive reference is: 
https://fuckingclosuresyntax.com/

You might also find this guide useful: https://www.hackingwithswift.com/example-code/language/what-is-trailing-closure-syntax

PR #126, opened 2020-03-03
(1/n) Provide conversion code and testing for `Currency` protocol buffer

keefertaylor:

Provide a native swift wrapper for the [Currency protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41). 

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan


XpringKit/Org_Xrpl_Rpc_V1_CurrencyAmount+XRPCurrency.swift
```diff
@@ -0,0 +1,6 @@
+internal extension XRPCurrency {
```
hbergren: What is an `internal extension` in Swiftland?

XpringKit/XRPCurrency.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// A currency on the XRP Ledger
```
hbergren: ???

```suggestion
/// An issued currency on the XRP Ledger
```

XpringKit/XRPCurrency.swift
```diff
@@ -0,0 +1,8 @@
...
+/// - SeeAlso: 
+public struct XRPCurrency: Equatable {
+  public let name: String
+  public let code: Data
+}
```
hbergren: What does the `Equatable` keyword mean here? Why do we need this `struct` and also the `internal extension`?

Tests/ProtocolBufferConversionTests.swift
```diff
@@ -0,0 +1,22 @@
+import XCTest
+@testable import XpringKit
+
+/// Tests conversion of protocol buffer to native Swift structs.
+final class ProtocolBufferConversionTests: XCTestCase {
+  func testConvertCurrency() {
+    // GIVEN a Currency protocol buffer with a code and a name.
+    let currencyCode = Data([1, 2, 3])
```
hbergren: What is `currencyCode` in the real-world? What does it signify?

XpringKit/Org_Xrpl_Rpc_V1_CurrencyAmount+XRPCurrency.swift
keefertaylor: Extension is adding functionality to this class. 

This is adding a new initializer to XRPCurrency that takes a protocol buffer. We could do this in one file but it's idiomatic in Swift to separate out into extensions when you're tightly coupling two classes.

XpringKit/XRPCurrency.swift
keefertaylor: `struct` is a class in swift, except it has a bunch of useful properties, like copy on assignment and a bunch of generated code under the hood (ie. for equality). Basically, they're like light classes. 
https://docs.swift.org/swift-book/LanguageGuide/ClassesAndStructures.html

`: Equatable` Means that we're conforming to the `Equatable` protocol. This basically means we implement an `equals` method, and then we can do things like `XCTAssertEquals(someCurrencyStruct, someOtherCurrencyStruct)`. `XCTAssertEquals` takes in two `Equatable` objects as opposed to an exact type:
https://developer.apple.com/documentation/swift/equatable

If we used a class we'd have to write our own equals method, but because we're using a `struct` and all of our iVars are also `equatable` the swift compiler automatically synthesizes the methods to make this class conform to `equatable` at runtime.

keefertaylor: I should have filled in the `SeeAlso` above. 

> The rippled APIs support two formats of currency code for issued currencies:
> - Standard Currency Codes: As a 3-character string such as "EUR" or "USD".
> - Nonstandard Currency Codes: As a 160-bit hexadecimal string, such as "0158415500000000C1F76FF6ECB0BAC600000000". This is uncommon.


https://xrpl.org/currency-formats.html#currency-codes

Tests/ProtocolBufferConversionTests.swift
keefertaylor: I should have filled in the `SeeAlso` above. 

> The rippled APIs support two formats of currency code for issued currencies:
> - Standard Currency Codes: As a 3-character string such as "EUR" or "USD".
> - Nonstandard Currency Codes: As a 160-bit hexadecimal string, such as "0158415500000000C1F76FF6ECB0BAC600000000". This is uncommon.


https://xrpl.org/currency-formats.html#currency-codes

PR #127, opened 2020-03-03
(2/n) Path / PathElement

keefertaylor:

Provide a native swift wrapper for the [Path Element protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L227) and [Path protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L237). 

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan


XpringKit/Org_Xrpl_Rpc_V1_Transaction.PathElement+XRPPathElement.swift
```diff
@@ -0,0 +1,7 @@
+internal extension XRPPathElement {
+  init(pathElement: Org_Xrpl_Rpc_V1_Payment.PathElement) {
+    self.account = pathElement.hasAccount ? pathElement.account.address : nil
+    self.currency = pathElement.hasCurrency ? XRPCurrency(currency: pathElement.currency) : nil
+    self.issuer = pathElement.hasIssuer ? pathElement.issuer.address : nil
+  }
+}
```
amiecorso: is it ok for these fields of XRPPathElement to be optional?  I see that in the proto none of account, currency or issuer are optional?

XpringKit/XRPCurrency.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// A currency on the XRP Ledger
+/// - SeeAlso: 
+public struct XRPCurrency: Equatable {
```
amiecorso: Is Equatable protocol like operator overloading for comparability?

XpringKit/XRPPath.swift
```diff
@@ -0,0 +1,3 @@
+public struct XRPPath: Equatable {
```
amiecorso: https://xrpl.org/paths.html
Paths, very cool.

XpringKit/XRPCurrency.swift
keefertaylor: It's a protocol that says we can use the `==` operator.  

It means we have to implement a `==` method (recall that in swift operators are functions), but in turn methods like `XCTAssertEquals` which takes `Equatable` objects magically works. 

You may see we haven't implemented the `==` method. That is because if we use a `struct` (instead of a class) and all our iVars are also `Equatable` then the compiler magically makes a `==` method for us.

XpringKit/Org_Xrpl_Rpc_V1_Transaction.PathElement+XRPPathElement.swift
keefertaylor: There are no required fields in proto3:
https://github.com/apple/swift-protobuf/blob/master/Documentation/API.md#type-modifiers

Instead there are default values:
https://github.com/apple/swift-protobuf/blob/master/Documentation/API.md#type-modifiers

As such, we check the `hasFoo` to decide if these should be set or not. 
-----

A good way to think about this is destinationTags. 
- If you didn't set a destination tag, then `.destinationTag` returns 0 (default value)
- If you set a destination tag to 0, then `.destinationTag` returns 0 (the set value)

These are useful to distinguish between.

PR #129, opened 2020-03-03
(4/n) Convert IssuedCurrency protocol buffer

keefertaylor:

Provide a native swift wrapper for the [IssuedCurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L28).

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan


XpringKit/Org_Xrpl_Rpc_V1_IssuedCurrencyAmount+XRPIssuedCurrency.swift
```diff
@@ -0,0 +1,13 @@
...
+
+internal extension XRPIssuedCurrency {
+  init?(issuedCurrency: Org_Xrpl_Rpc_V1_IssuedCurrencyAmount) {
+    guard let value = BigInt(issuedCurrency.value) else {
+      return nil
+    }
+
+    self.currency = XRPCurrency(currency: issuedCurrency.currency)
+    self.value = value
+    self.issuer = issuedCurrency.issuer.address
```
amiecorso: what would happen if issuedCurrency.issuer didn't exist or didn't have an address?  Is this impossible?

keefertaylor: It's impossible based on API constraints. This proto will just default to the empty string if it didn't exist.

PR #131, opened 2020-03-03
(6/n) Convert Payment Protocol Buffers

keefertaylor:

Provide a native swift wrapper for the [Payment protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L224). 

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan


Tests/ProtocolBufferConversionTests.swift
```diff
@@ -207,4 +202,174 @@ final class ProtocolBufferConversionTests: XCTestCase {
...
+    XCTAssertEqual(payment?.destination, paymentProto.destination.value.address)
+    XCTAssertEqual(payment?.destinationTag, paymentProto.destinationTag.value)
+    XCTAssertEqual(payment?.deliverMin, XRPCurrencyAmount(currencyAmount: paymentProto.deliverMin.value))
+    XCTAssertEqual(payment?.invoiceID, paymentProto.invoiceID.value)
+    XCTAssertEqual(payment?.paths, paymentProto.paths.map { XRPPath(path: $0) })
+    XCTAssertEqual(payment?.sendMax, XRPCurrencyAmount(currencyAmount: paymentProto.sendMax.value))
+  }
+
+  func testConvertPaymentWithOnlyMandatoryFieldsSet() {
+    // GIVEN a pyament protocol buffer with only mandatory fields set.
```
hbergren:
```suggestion
    // GIVEN a payment protocol buffer with only mandatory fields set.
```

https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

Tests/ProtocolBufferConversionTests.swift
```diff
@@ -207,4 +202,174 @@ final class ProtocolBufferConversionTests: XCTestCase {
...
+        }
+      }
+    }
+
+    // WHEN the protocol buffer is converted to a native Swift type THEN the result is nil
+    XCTAssertNil(XRPPayment(payment: paymentProto))
+  }
+
+  func testConvertPaymentWithInvalidDeliverMinField() {
+    // GIVEN a pyament protocol buffer with an invalid deliverMin field
```
hbergren:
```suggestion
    // GIVEN a payment protocol buffer with an invalid deliverMin field
```

Tests/ProtocolBufferConversionTests.swift
```diff
@@ -207,4 +202,174 @@ final class ProtocolBufferConversionTests: XCTestCase {
...
+        }
+      }
+    }
+
+    // WHEN the protocol buffer is converted to a native Swift type THEN the result is nil
+    XCTAssertNil(XRPPayment(payment: paymentProto))
+  }
+
+  func testConvertPaymentWithInvalidSendMaxField() {
+    // GIVEN a pyament protocol buffer with an invalid sendMax field
```
hbergren:
```suggestion
    // GIVEN a payment protocol buffer with an invalid sendMax field
```

PR #134, opened 2020-03-05
Renames XpringClient to XRPClient

amiecorso:
XRPClient is a more appropriate name for the client class in Xpring SDKs. This PR begins the process of updating this name while maintaining the current implementation of XpringClient for gradual deprecation.

### Context of Change
This PR updates all "XpringClient" related names in the SDK to "XRPClient", such as "DefaultXpringClient" --> "DefaultXRPClient", as well as local variables such as "xpringClient" --> "xrpClient". It also introduces a new top-level client class "XRPClient" while maintaining the original "XpringClient" class for eventual deprecation. The "XpringClient" class now wraps an instance of "DefaultXRPClient".

Analogues in other SDKs:
https://github.com/xpring-eng/xpring4j/pull/113
https://github.com/xpring-eng/Xpring-JS/pull/203

### Type of Change

## Before / After
Before: Everything named as "XpringClient"
After: top level "XpringClient" still exists, but everything else is now "XRPClient", including identical top-level class by this name.

## Test Plan
CircleCI

Future tasks include updating names correspondingly in the other SDKs, and eventually deprecating the XpringClient class.


keefertaylor:
meta nit: Use \` in your PR descriptions for readability

Tests/XpringClientIntegrationTests.swift
```diff
@@ -1,6 +1,7 @@
 import XCTest
 import XpringKit
 
+/*
```
keefertaylor: extensions occur globally so these two things are going to trample on each other. If this wasn't test code we should probably name these things more generically and / or have them in their own files. Since this test file will go away soon, I think it's fine to just rely on them being added in `XRPClientIntegrationTests`.  Thus, you can delete this block comment.

keefertaylor:
I think your test is crashing here because you're bundling an old version of xpring-common-js which doesn't have the PayID functions we need.

hbergren:
@amiecorso , can you give me a ping to review when conflicts are resolved? (And tests pass?)

Tests/XpringClientIntegrationTests.swift
```diff
@@ -0,0 +1,78 @@
...
+/// Integration tests run against a live remote client.
+final class XpringClientIntegrationTests: XCTestCase {
+  private let legacyClient = XpringClient(grpcURL: .legacyRemoteURL, useNewProtocolBuffers: false)
+  private let client = XpringClient(grpcURL: .remoteURL, useNewProtocolBuffers: true)
+
+  // MARK: - rippled Protocol Buffers
+
+  func testGetBalance() {
+    do {
+      _ = try client.getBalance(for: Wallet.testWallet.address)
```
hbergren: Where is `Wallet.testWallet` in scope here? is that from `import XpringKit`?

keefertaylor: This comes from this file:
https://github.com/xpring-eng/XpringKit/blob/master/Tests/Helpers/Wallet%2BTest.swift

We're using extensions to add a static property onto the `Wallet` object. This lives in the `XpringKitTests` target. This pattern looks weird at first but it's really useful to be able to attach static properties onto classes under test from XpringKit. In fact, we have an entire folder of them: 
https://github.com/xpring-eng/XpringKit/tree/master/Tests/Helpers

PR #144, opened 2020-03-09
Add Fakes for `URLSession`, `URLSessionDataTask`

keefertaylor:

Add fakes for built in networking classes in Swift. 

### Context of Change

Swift has built in classes for networking. If we fake them then we can create hermetic unit tests. 

PayID is going to use HTTP to interact with remote PayID services. In order to test this we'll need these fakes. They are split into a separate logical PR for reviewers ease and will get utilized in #143.

Generally took this code from https://www.swiftbysundell.com/articles/mocking-in-swift/ with minor alterations for naming and data types. 

### Type of Change

## Before / After

Can easily test.

## Test Plan

N/A

## Future Tasks

At some point we should evaluate whether it would make sense to use a networking library like AFNetworking or AlamoFire rather than the built in classes. Since things are still relatively simple, I'm not over engineering in this PR.

### DIFF
--- a/Tests/Fakes/FakeURLSession.swift
+++ b/Tests/Fakes/FakeURLSession.swift
@@ -1,20 +1,44 @@
 import Foundation
 
-/// Fakes a URLSession for testing.
-class FakeURLSession: URLSession {
-  typealias CompletionHandler = (Data?, URLResponse?, Error?) -> Void
+/// A fake URLSession that will return data tasks which will call completion handlers with the given parameters.
+public class FakeURLSession: URLSession {
+  /// Fields that will be returned in the next URLSessionDataTask.
+  public var urlResponse: URLResponse?
+  public var data: Data?
+  public var error: Error?
 
-  // The data task that will be returned.
-  var fakeDataTask: FakeURLSessionDataTask
-
-  public init(fakeDataTask: FakeURLSessionDataTask) {
-    self.fakeDataTask = fakeDataTask
+  /// Initialize a new FakeURLSession.
+  ///
+  /// - Parameters:
+  ///   - string: A utf-8 encoded string that will be encoded as data in the next request.
+  ///   - urlResponse: The url response to serve in the next request.
+  ///   - error: The error to serve in the next request.
+  public convenience init(string: String, urlResponse: HTTPURLResponse?, error: Error?) {
+    let data = string.data(using: .utf8)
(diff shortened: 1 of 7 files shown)


Tests/Fakes/FakeURLSession.swift
```diff
@@ -0,0 +1,44 @@
...
+  public var data: Data?
+  public var error: Error?
+
+  /// Initialize a new FakeURLSession.
+  ///
+  /// - Parameters:
+  ///   - string: A utf-8 encoded string that will be encoded as data in the next request.
+  ///   - urlResponse: The url response to serve in the next request.
+  ///   - error: The error to serve in the next request.
+  public convenience init(string: String, urlResponse: HTTPURLResponse?, error: Error?) {
```
hbergren: What's this `convenience` init?

An override if the first parameter is `string`?

keefertaylor: `convenience` means that this initializer will call another initializer. We call this a convenience initializer. We call the initializer who does't have `convenience` in it a 'designated initializer`. 

Convenience initializers:
- cannot call super class initializers
- must call a designated initializer

Designated Initializers:
- cannot call convenience initializer
- can call a super class initializer

-----

Why is this a convenience initializer? 

The first parameter we want is a `Data` object. `Data` is just a wrapper classer for a collection of bytes, so it can be virtually anything. 

However, since most data we get back from web requests is just utf-8 encoded text and clients likely want to to work with text rather than bytes, this convenience initializer knows how to transform a string into bytes and serve that to a designated intializer.

The effect of this is that we write:
```
    let data = string.data(using: .utf8)
```
once in our class, rather than all of the clients of this class having to write this code every time they want to fake a string response.

keefertaylor:
- Austin / + Amie since Amie has Swift context and Austin is busy this week.

Tests/Fakes/FakeURLSession.swift
hbergren: Great explanation. Thanks for the Swift tutorial!

keefertaylor:
- Austin who is slammed this week, + Amie

Tests/Fakes/FakeURLSession.swift
amiecorso: yeah this is gold

keefertaylor:
Closing. This got merged in #143

PR #156, opened 2020-03-11
Protobuf Wrapper Objects (1/3)

nkramer44:
Asana task: https://app.asana.com/0/1143722888069470/1165556580466065

- Added data classes that will get converted to/from protobuf objects to avoid exposing protobufs in the public API

### Context of Change
See https://github.com/xpring-eng/Xpring-JS/pull/246 for more detail.

This is just the swift counterpart to what we've done in Xpring4j and what is in flight for Xpring-JS

### Type of Change

## Before / After
Before: `IlpClient` consumes and produces protobuf objects
After: `IlpClient` still consumes and produces protobuf, but in part 2 of this PR, it will consume and produce these new wrapper objects.

## Test Plan
The existing test suite will suffice for this PR, but will need some refactoring (ie changing request/response types) once these objects are in use.

## Future Tasks

- Change `IlpClient` signatures and refactor tests to use wrapped objects

### DIFF
--- a/XpringKit/ILP/model/AccountBalance.swift
+++ b/XpringKit/ILP/model/AccountBalance.swift
@@ -1,22 +1,19 @@
 /// Response object for requests to get an account's balance
 public struct AccountBalance {
-    
+
     /// The accountId for this account balance.
     public let accountID: AccountID
-    
-    
+
     /// Currency code or other asset identifier that this account's balances will be denominated in
     public let assetCode: String
-    
-    
+
     /// Interledger amounts are integers, but most currencies are typically represented as # fractional units, e.g. cents.
     /// This property defines how many Interledger units make # up one regular unit.
     /// For dollars, this would usually be set to 9, so that Interledger # amounts are expressed in nano-dollars.
     ///
     /// This is an Int32 representing this account's asset scale.
     public let assetScale: Int32
-    
-    
+
     /// The amount of units representing the clearing position this Connector operator holds with the account owner.
     /// A positive clearing balance indicates the Connector operator has an outstanding liability
     /// (i.e., owes money) to the account holder.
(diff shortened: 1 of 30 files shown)


XpringKit/ILP/model/AccountBalance.swift
```diff
@@ -0,0 +1,73 @@
...
+    /// liability (i.e., owes money) to the account holder.
+    /// A negative balance represents an asset (i.e., the account holder owes money to the operator).
+    /// This value is the sum of the clearing balance and the prepaid amount.
+    ///
+    /// A Int64 representing the net clearingBalance of this account.
+    public let netBalance: Int64
+
+    /// Private constructor to initialize an AccountBalance.
+    /// Explicitly declare a constructor to derive netBalance
+    private init(
```
keefertaylor: Your pattern works and there is some precedent to use it because it's the same in JS and Java. 

Because you are using a struct, if you leave off any initializer the compiler will automatically generate you one with the `internal` access level. `internal` is "private to the class, subclasses, package and tests" which is a sane default. 

Generally in Swift, static factory methods aren't particularly common. Not necessarily an anti-pattern. 

You can get an initializer auto generated if you use an extension. For instance, in `AccountBalance+Org_Interledger_Stream_Proto_GetBalanceResponse.swift` you could write:

```
extension AccountBalance {
   public convenience init(getBalanceResponse: Org_Interledger_Stream_Proto_GetBalanceResponse) {
        self.init(accountId: getBalanceResponse.getAccountID(), ...)_
   }
}
```

This is useful for two reasons:
1) It keeps class specific logic out of the core struct, which encourages proper encapsulation / decoupling
2) because you haven't written an `init` in the struct you still get the autogenerated one

A few notes:
- If you keep the static factory methods, that's fine. A more idiomatic name would be `make` rather than `from`. 
- If you do the extension thing, I named my files backwards (ie. it should be `AccountBalance+Org_Interledger_Stream_Proto_GetBalanceResponse.swift` rather than `Org_Interledger_Stream_Proto_GetBalanceResponse+AccountBalance.swift`). The first class name is what's being extended.

nkramer44: I have no problem with the extension constructor over a static factory.  It seems like a more swift-y way of doing what I did in Java and JS, which is a good thing.

I think I'm grokking what you're suggesting, but I'll take another pass at it and re-request your review when I do.

nkramer44: Just to confirm, the extensions of model objects don't actually extend protobuf objects right? They just add a constructor with a proto object? 

If so, does that file naming convention still apply, or is that usually reserved for an object extending another object?

PR #160, opened 2020-03-16
Protobuf wrapper objects (2+3/3)

nkramer44:
- Change `getBalance` and `sendPayment` signatures to use proto wrapper objects

### Context of Change
Swift counterpart to https://github.com/xpring-eng/xpring4j/pull/126 and https://github.com/xpring-eng/xpring4j/pull/127.

Putting both of these changes into a singular PR for the sake of velocity, and because this basically mirrors what we did in Xpring4j

### Type of Change

## Before / After
Before: `getBalance` and `sendPayment` consumed and returned proto objects
After: `getBalance` and `sendPayment` consume and return wrapper objects

## Test Plan
Updated `DefaultIlpClientTest` to check each field of responses, which expands the scope of those tests to the conversion process between protos and non-protos


Tests/DefaultIlpClientTest.swift
```diff
@@ -13,7 +13,7 @@ final class DefaultIlpClientTest: XCTestCase {
         )
 
         // WHEN the balance is requested
-        guard let balance: Org_Interledger_Stream_Proto_GetBalanceResponse =
+        guard let balance: AccountBalance =
```
keefertaylor: you actually don't need to type this. You can just let the type be inferred with:
`guard let balance = try? ...`

PR #162, opened 2020-03-18
Wrap Metadata in Network Calls

nkramer44:
Asana task: https://app.asana.com/0/1154633712169520/1165556580466061
- Introduce a wrapper around `Metadata` class, which is passed to ILP network client calls to provide authentication to the connector
    - The wrapper ensures that bearer tokens are always sent over the wire with a `Bearer ` prefix

### Context of Change
Currently in `DefaultIlpClient`, we are manually constructing a `Metadata` object and adding an `authorization` entry before making network client calls. 

However, the SDK doesn't enforce that the outgoing token is prefixed with Bearer . This does not break anything, because Hermes/java connectors are smart enough to detect when there is no prefix, but the Bearer prefix standard should be followed by the client regardless.

Unlike this change in Xpring-JS (https://github.com/xpring-eng/Xpring-JS/pull/267), you cannot extend `Metadata` because it is not `open`.  Though it's not as pretty, `IlpCredentials` here just wraps `Metadata` instead of extending it.

### Type of Change

## Before / After
Before: `Metadata` constructed in network clients and `Authorization` added manually, with our without `Bearer ` prefix
After: `Metadata` is replaced with `IlpCredentials`, which enforces a `Bearer ` prefix in the `Authorization` header that it adds.

## Test Plan
Added a unit test class to test the `IlpCredentials` constructor.

### DIFF
--- a/XpringKit/ILP/auth/IlpCredentials.swift
+++ b/XpringKit/ILP/auth/IlpCredentials.swift
@@ -14,7 +14,7 @@ class IlpCredentials {
     /// of that entry will be the bearerToken with "Bearer " prefix.
     public init(_ bearerToken: String) throws {
         self.metadata = Metadata()
-        try metadata.add(key: "Authorization", value: self.applyBearer(bearerToken))
+        try metadata.add(key: "authorization", value: self.applyBearer(bearerToken))
     }
 
     /// Prepends 'Bearer ' to an auth token, if it is not already prefixed with "Bearer "
--- a/XpringKit/ILP/auth/IlpCredentials.swift
+++ b/XpringKit/ILP/auth/IlpCredentials.swift
@@ -3,7 +3,7 @@ import SwiftGRPC
 /// A helper which provides a convenient way to initialize grpc.Metadata and
 /// add an Authorization metadata header, and ensures every bearer token
 /// going over the wire is prefixed with "Bearer "
-class IlpCredentials {
+internal class IlpCredentials {
 
     /// Unfortunately, Metadata is not an open class in SwiftGRPC, so we cannot extend it.
     /// Instead, it is wrapped in IlpCredentials
@@ -27,7 +27,7 @@ class IlpCredentials {
     }
 
     /// Get the Metadata to pass to network calls
-    public func get() -> Metadata {
+    public func getMetadata() -> Metadata {
         return self.metadata
     }
 }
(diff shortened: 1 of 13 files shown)


XpringKit/ILP/auth/IlpCredentials.swift
```diff
@@ -0,0 +1,33 @@
+import SwiftGRPC
+
+/// A helper which provides a convenient way to initialize grpc.Metadata and
+/// add an Authorization metadata header, and ensures every bearer token
+/// going over the wire is prefixed with "Bearer "
+class IlpCredentials {
+
+    /// Unfortunately, Metadata is not an open class in SwiftGRPC, so we cannot extend it.
```
keefertaylor: bummer

XpringKit/ILP/auth/IlpCredentials.swift
```diff
@@ -0,0 +1,33 @@
+import SwiftGRPC
+
+/// A helper which provides a convenient way to initialize grpc.Metadata and
+/// add an Authorization metadata header, and ensures every bearer token
+/// going over the wire is prefixed with "Bearer "
+class IlpCredentials {
```
keefertaylor: Any reason to expose this? Should we label this as internal?

XpringKit/ILP/auth/IlpCredentials.swift
```diff
@@ -0,0 +1,33 @@
...
+    ///
+    /// - Parameters:
+    ///     - token: An auth token that either does or does not have a "Bearer " prefix
+    /// - Returns: A bearer token as a String, with a "Bearer " prefix
+    private func applyBearer(_ token: String) -> String {
+        return token.starts(with: "Bearer ") ? token : "Bearer " + token
+    }
+
+    /// Get the Metadata to pass to network calls
+    public func get() -> Metadata {
```
keefertaylor: Is `MetaData` a struct or a class? If it's a struct then it's copy on assign and you can just make your variable public. 

If it isn't I'd recommend using `getMetadata` or something? `get()` is only descriptive if you know the class structure, IMO

XpringKit/ILP/auth/IlpCredentials.swift
nkramer44: No, yes

XpringKit/ILP/auth/IlpCredentials.swift
nkramer44: `Metadata` is a class.  I considered doing something like `getMetadata`, but ended up going with `get()` since `metadata` is the only thing we're wrapping here and `get()` just rolls off the tongue a little better.  I'll change to `getMetadata` though, since it's clearer and I'm not super worried about being pretty in internal stuff.

keefertaylor: wow, bummer on the class. This sounds reasonable though.

Issue #51, opened 2019-12-26
Error Installing Cocoa Pods


dangell7:
Error installing cocoa pods on swift 5

```
[!] /bin/bash -c 
set -e
git submodule update --init --recursive && ./scripts/bundle_js.sh && ./scripts/regenerate_protos.sh

Submodule 'xpring-common-js' (git@github.com:xpring-eng/xpring-common-js) registered for path 'xpring-common-js'
Submodule 'xpring-common-protocol-buffers' (git@github.com:xpring-eng/xpring-common-protocol-buffers) registered for path 'xpring-common-protocol-buffers'
Cloning into '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-js'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:xpring-eng/xpring-common-js' into submodule path '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-js' failed
Failed to clone 'xpring-common-js'. Retry scheduled
Cloning into '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-protocol-buffers'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:xpring-eng/xpring-common-protocol-buffers' into submodule path '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-protocol-buffers' failed
Failed to clone 'xpring-common-protocol-buffers'. Retry scheduled
Cloning into '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-js'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:xpring-eng/xpring-common-js' into submodule path '/Users/denisangell/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.1.3-42df3/xpring-common-js' failed
Failed to clone 'xpring-common-js' a second time, aborting
```

[...]

Issue #44, opened 2019-12-12
Failing to install XpringKit with Cocoapods


doniekelly:
I'm a bit lost as to why this is failing as nothing is obvious to me as to what is the issue. Any help appreciated. 

Thanks. Logs below after I run pod update 

```
Mac:XRPL donie$ pod update
Update all pods
Updating local specs repositories
Analyzing dependencies
Downloading dependencies
Installing BigInt (5.0.0)
Installing BoringSSL-GRPC (0.0.3)
Installing SwiftGRPC (0.10.0)
Installing SwiftProtobuf (1.7.0)
Installing XpringKit (1.2.0)
Enter passphrase for key '/Users/donie/.ssh/id_rsa': 
Enter passphrase for key '/Users/donie/.ssh/id_rsa': 

[!] /bin/bash -c 
set -e
git submodule update --init --recursive && ./scripts/bundle_js.sh && ./scripts/regenerate_protos.sh

Submodule path 'xpring-common-js': checked out '4ce627e8ed33f44e9870fd84695a7d7c18f80884'
Submodule path 'xpring-common-js/xpring-common-protocol-buffers': checked out '3603db829677c41a7789746b858feb19a6e77845'
Submodule path 'xpring-common-protocol-buffers': checked out '41d8d7c316d11b79ab495110da24b9fb0f2c691b'
Bundling JS
>> Installing Node Dependencies

> fsevents@1.2.9 install /Users/donie/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.2.0-83e16/xpring-common-js/node_modules/fsevents
> node install

  SOLINK_MODULE(target) Release/.node
  CXX(target) Release/obj.target/fse/fsevents.o
  SOLINK_MODULE(target) Release/fse.node
  COPY /Users/donie/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.2.0-83e16/xpring-common-js/node_modules/fsevents/lib/binding/Release/node-v79-darwin-x64/fse.node
  TOUCH Release/obj.target/action_after_build.stamp

> tiny-secp256k1@1.1.3 install /Users/donie/Library/Caches/CocoaPods/Pods/Release/XpringKit/1.2.0-83e16/xpring-common-js/node_modules/tiny-secp256k1
> npm run build || echo "secp256k1 bindings compilation fail. Pure JS implementation will be used."


[...]

PR #167, opened 2020-03-24
PayID Get Via Swagger Codegeneration

keefertaylor:

Generate and implement a PayID API via Swagger

### Context of Change

Implement:
- Swagger code generation, associated artifacts and scripts
- `PayIDClient` to resolve XRP Addresses
- Integration tests

Unit tests are omitted because the generated code is hard to test. Proper testing tracked in https://app.asana.com/0/1143722888069470/1170642421030123. Given that we have integration tests I am confident in the code working and I feel it is pragmatic to land this now. 

Recall that code in `Generated` is auto-generated and not for review. We check it in for Carthage compatibility. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

PayID exists in Swift

## Test Plan

New integration tests provided for CI


keefertaylor:
I'm not sure what specifically changed that updated the spacing in the .proto files. Apologies for the noise in this review.

pay-id-api-spec/pay-id.v1.yml
```diff
@@ -0,0 +1,333 @@
...
+        - address
+    PaymentInformation:
+      title: PaymentInformation
+      type: object
+      properties:
+        addressDetailsType:
+          type: string
+        addressDetails:
+          $ref: '#/components/schemas/CryptoAddressDetails'
+        proof_of_control_signature:
```
austin-king, line 159: Does this allow us to indicate optional fields?

keefertaylor: Yup, you'll see below (line 163) that there's a required field set. This field is optional because it's not in the required fields.

austin-king: got it, thanks

keefertaylor:
> Includes generated code, should be be ignoring that directory?

Unfortunately, the Carthage build system checks out our Swift code from GitHub and builds it. If we force folks to generate code, then they have to have a bunch of extra tools installed and this fails in very hard to track down ways and has been the cause of a few issues (Specifically #44  and #51 )

PR #174, opened 2020-03-27
Adds xrpl.org link to XRPPayment class

amiecorso:
All of our language-native protobuf wrapper classes should link to the documentation of the corresponding XRPL concept on xrpl.org.  We forgot this in the `XRPPayment` class.
(Also taken care of in Swift and JS)

### Type of Change

- [x] Documentation Updates

### DIFF
--- a/XpringKit/XRPPayment.swift
+++ b/XpringKit/XRPPayment.swift
@@ -1,8 +1,8 @@
 import Foundation
 
-// TODO(keefertaylor): Modify this object to use X-Address format.
 /// A payment on the XRP Ledger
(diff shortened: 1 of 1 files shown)


XpringKit/XRPPayment.swift
```diff
@@ -1,6 +1,8 @@
 import Foundation
 
 // TODO(keefertaylor): Modify this object to use X-Address format.
```
keefertaylor: Can you place this TODO after the class comment? 

So it reads:
<summary>
<random TODOs>
<code>

PR #175, opened 2020-03-27
Translate gRPC Errors to XpringIlpErrors

nkramer44:
**Sorry for long PR.  Most of it is fakes/code that looks very similar in tests**

Translate and throw errors we get from gRPC network calls to `XpringIlpError`s

### Context of Change
Errors that we get back from gRPC services aren't all that informative or easy to work with.  We also want to hide the network layer from end users, and rethrowing gRPC errors doesn't accomplish this.  By translating gRPC errors into native `XpringIlpError`s, we have more control over what the end user is dealing with and can provide a richer error model for them.

### Type of Change

Doubt anyone is catching gRPC errors explicitly, but if they are, this would break that logic.
## Before / After
`IlpClient.getBalance` and `IlpClient.sendPayment` would return `RPCError` if something went wrong
Now they throw `XpringIlpError`s

## Test Plan
Added a ton of fake network error responses and test cases to `DefaultIlpClientTests` for negative cases


amiecorso:
Note to selves:  Xpring-JS Analogue: https://github.com/xpring-eng/Xpring-JS/pull/330#pullrequestreview-383317879

Tests/Helpers/FakeIlpPaymentNetworkClient+Test.swift
```diff
@@ -1,8 +1,34 @@
...
-    static let successfulFakeNetworkClient = FakeIlpPaymentNetworkClient(
-        sendPaymentResult: .success(.testSendPaymentResponse)
-    )
+    static let successfulFakeNetworkClient = FakeIlpPaymentNetworkClient(sendPaymentResult: .success(.testSendPaymentResponse))
+
+    /// Creates a FakeIlpPaymentNetworkClient which will always throw the specified Error
+    ///
+    /// - Parameters:
+    ///     - errorResponse: An Error that will be thrown by the network client
+    public static func withErrorResponse(_ errorResponse: Error) -> FakeIlpPaymentNetworkClient {
```
keefertaylor, line 12: optional nit: I think it's slightly more idiomatic to call this:
```
with(errorResponse: Error) 
```

But I can't explicitly find anything in the Swift naming guide on this:
https://swift.org/documentation/api-design-guidelines/

But I can't

XpringKit/ILP/IlpError.swift
```diff
@@ -0,0 +1,48 @@
+import SwiftGRPC
+
+/// Errors that may occur when interacting with Xpring's ILP infrastructure
+public enum IlpError: Error {
+
+    /// The requested functionality is not yet implemented.
```
keefertaylor, line 6: Alphabetize your cases.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Changed
 - `XRPClient` now uses [rippled's protocol buffer API](https://github.com/ripple/rippled/pull/3254) rather than the legacy API. Users who wish to use the legacy API should pass `false` for `useNewProtocolBuffers` in the constructor.
+- `IlpClient` methods now throw `XpringIlpError`s if something goes wrong during the call (either client side or server side).  This is only breaking if users are handling special error cases, which were previously `RPCError`s
```
keefertaylor, line 12: s/XpringIlpError/IlpError

PR #191, opened 2020-04-10
New PayID Format Support

keefertaylor:

Migrate PayID format from:
`$xpring.money/georgewashington`
to
`georgewashington$xpring.money`

### Context of Change

Product requested change in format. 

Remove references to 'payment pointer' and rename a few classes / methods to reflect that we're not longer compliant with the payment pointer spec.

This is a breaking change, but PayID is experimental so I don't consider it a big deal.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

New spec implemented.

## Test Plan

New tests provided for CI. 

## Future Tasks
Uptake parsing logic and test cases in client libraries (JS / Swift / Java)


intelliot:
This file may need to be updated or removed https://github.com/xpring-eng/XpringKit/blob/ea927120c7cf0fbdbab7330a1f81a22122cfba8d/XpringKit/PayID/PaymentPointer.swift

keefertaylor:
> This file may need to be updated or removed
ILP still uses this file so I moved it to their directory.

Tests/PayID/PayIDUtilsTests.swift
```diff
@@ -2,63 +2,67 @@ import XCTest
...
 
 class PayIDUtilsTest: XCTestCase {
-  func testParsePaymentPointerHostAndPath() {
-    // GIVEN a payment pointer with a host and a path.
-    let rawPaymentPointer = "$example.com/foo"
+  func testParseValidPayID() {
+    // GIVEN a Pay ID with a host and a path.
+    let host = "xpring.money"
+    let path = "georgewashington"
+    let rawPayID = "\(path)$\(host)"
```
hbergren, line 9: The variable names obscure this a little, so I just want to make sure we're on the same page about the format here.

`alice$xpring.money` => `https://xpring.money/alice`

`alice$xpring.money/example-path` => `https://xpring.money/example-path/alice`.

The format here is `{user}${domain.tld(/namespace-path)}`

Tests/PayID/PayIDUtilsTests.swift
```diff
@@ -2,63 +2,67 @@ import XCTest
...
-    XCTAssertEqual(paymentPointer?.path, PaymentPointer.paymentPointerWellKnownPath)
+    // THEN the Pay ID failed to parse.
+    XCTAssertNil(payIDComponents)
   }
 
-  func testParsePaymentPointerIncorrectPrefix() {
-    // GIVEN a payment pointer without a '$' prefix
-    let rawPaymentPointer = "example.com/"
+  func testParsePayIDEmptyPath() {
+    // GIVEN a Pay ID with an empty host.
```
hbergren, line 46:
```suggestion
    // GIVEN a Pay ID with an empty user.
```

Tests/PayID/PayIDUtilsTests.swift
keefertaylor: This is the first I'm hearing of supporting arbitrary paths. Given that no one has mentioned this, and I can't find it located anywhere else, I'm going to ignore this edge case for now. If/when we standardize on this, I'm happy to make sure it's supported with unit tests. 

I think that this will just work; but without a proper specification from product or engineering there's really no way to know.

hbergren: Ok, the executive decision is that if you want to namespace, use a subdomain, so you don't have to worry about the `path` stuff.

Tests/PayID/PayIDUtilsTests.swift
```diff
@@ -2,63 +2,67 @@ import XCTest
...
+    // THEN the Pay ID failed to parse.
+    XCTAssertNil(payIDComponents)
   }
 
-  func testParsePaymentPointerIncorrectPrefix() {
-    // GIVEN a payment pointer without a '$' prefix
-    let rawPaymentPointer = "example.com/"
+  func testParsePayIDEmptyPath() {
+    // GIVEN a Pay ID with an empty user.
+    let host = "xpring.money" // Extra '$'
```
austin-king, line 47: What do you mean by "Extra '$'"?

Tests/PayID/PayIDUtilsTests.swift
```diff
@@ -2,63 +2,67 @@ import XCTest
...
   }
 
-  func testParsePaymentPointerWithWellKnownPath() {
-    // GIVEN a payment pointer with a well known path.
-    let rawPaymentPointer = "$example.com"
+  func testParsePayIDTooManyDollarSigns() {
+    // GIVEN a Pay ID with too many '$'.
+    let host = "xpring$money" // Extra '$'
+    let path = "georgewashington"
+    let rawPayID = "\(path)$\(host)"
```
nkramer44, line 23: I believe PayIDs can have as many `$`s as they want and still be a valid PayID (see https://github.com/xpring-eng/rfcs/blob/master/payid/src/spec/payid-uri.md#definition).  

> Another scenario is a payment service provider (e.g., a digital wallet) that allows a user to use a PayID that is associated with some other payment service provider. For example, a user with the PayID "alice$bank.example.net" might register with a wallet website whose domain name is "wallet.example.net". In order to use her bank's PayID as the localpart of the wallet's 'payid' URI, the dollar-sign character (U+0024) needs to be percent-encoded as described in [RFC3986]. Thus, the resulting 'payid' URI would be "payid:alice%24bank.example$wallet.example".

Here, the PayID `payid:georgewashington$xpring$money` should be parsed to `payid:georgewashington%24xpring$money`.

Not suggesting you need to implement all of this, and likely no one will ask for something like that in the near future, but it's something to consider.

XpringKit/PayID/PayIDUtils.swift
```diff
@@ -3,11 +3,11 @@ public enum PayIDUtils {
...
-  /// Parse a payment pointer string to a payment pointer object
+  /// Parse the given Pay ID to a set of components.
   ///
-  /// - Parameter paymentPointer: The input string payment pointer.
-  /// - Returns: A  `PaymentPointer` object if the input was valid, otherwise undefined.
-  public static func parse(paymentPointer: String) -> (host: String, path: String)? {
-    return javaScriptPayIDUtils.parse(paymentPointer: paymentPointer)
+  /// - Parameter payID: A PayID to parse.
+  /// - Returns: A set of components parsed from the PayID.
+  public static func parse(payID: String) -> (host: String, path: String)? {
```
nkramer44, line 10: To conform to the PayID-URI spec, this should return something like this:
`(scheme: String, host: String, account: String)`.

For example, a call to `parse(payID: "payid:nkramer$xpring.money")` should return 
`(scheme: "payid", host: "xpring.money", account: "nkramer")`

keefertaylor:
For now, I'm going to leave this as-is. AFAIK, we've neither presented nor fully spec'ed any of the other proposals and I don't think this change necessarily stops us from implementing that. 

Additionally, this logic is already live in xpring-common-js and xpring-js, so at least we'll become consistent.

Tests/PayID/PayIDUtilsTests.swift
keefertaylor: This is errant from above (the too many dollar signs test)

PR #192, opened 2020-04-10
Introduce a protocol for XRPClient

keefertaylor:

Introduce `XRPClientProtocol` and conform `XRPClient` and `FakeXRPClient` to the protocol. This will allow us to fake `XRPClient`s in the Xpring Component. 

### Context of Change

A future component, Xpring, will compose interactions between PayID and XRP. In order to test this component, we need to be able to fake interactions with the XRP Ledger. 

This is a prefactor to make that possible. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change

CI


XpringKit/XRP/XRPClientProtocol.swift
```diff
@@ -0,0 +1,47 @@
+/// An interface into the XRP Platform.
+public protocol XRPClientProtocol {
```
hbergren, line 2: So just to put this into TypeScript terms (which I understand), this defines an `interface` with basically `abstract` methods that both `XRPClient` and `FakeXRPClient` implement.

When you later test the `XpringClient`, you can create a class/instance that implements this interface, but returns known values so you can pass your test assertions, correct?

keefertaylor: Yup. `protocol` in swift == `interface` in typescript. Building this here lets us fake this later.

keefertaylor:
> Looks great. So this is just completing the addition of an XRPClient interface/protocol like what you've already done in JS, but keeping it consistent across SDKs as the need for creating fake XRPClients arises with testing XpringClient

Correct.

PR #194, opened 2020-04-10
Add a `FakeWallet`

keefertaylor:

Create a `FakeWallet` which always produces the given signature.

### Context of Change

This is needed to properly unit test some of the upcoming Xpring functionality.

Note that I only need to partially fake some of the functionality in this class, so I opted to extend `Wallet` and subclass, rather than writing a `Wallet` interface and perfectly hermetic fake. This is kind of lazy, but also pragmatic and we can always refactor later if needed. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A
## Test Plan

CI


keefertaylor:
> Nice, looks good to me. So the signature property of this FakeWallet isn't necessarily related to the provided public and private keys, right? Because it is set arbitrarily upon construction. We just need some keys to provide for successful construction of the superclass?

Yeah. They're provided because I'm taking a shortcut and inheriting rather than writing a full on protocol.

PR #195, opened 2020-04-10
Refactor `FakeXRPClient` to use `Result`s

keefertaylor:

Refactor `XRPClient` to use `Result` types rather than always successfully returning values. 

### Context of Change

In some future tests, I want to test `XRPClient` interactions when errors are thrown. This is a prefactor to make that possible.

This is a breaking change, but the broken code is only in tests.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

no change

## Test Plan

CI


hbergren:
Assuming CI passes.

PR #196, opened 2020-04-11
Adds documentation to protobuf wrapper classes (XRP component)

amiecorso:
Adds documentation based on the information in xrpl.org docs to the fields and constructors in native Swift protobuf wrapper classes. 

### Context of Change
The Java SDK is well documented in this capacity - JS and Swift need to be made consistent. 

### Type of Change

- [x] Documentation Updates

## Test Plan
None

## Future Tasks

Other two SDKs done

### DIFF
--- a/XpringKit/XRPPathElement.swift
+++ b/XpringKit/XRPPathElement.swift
@@ -1,7 +1,20 @@
 /// A path on the XRPLedger
 /// - SeeAlso: https://xrpl.org/paths.html#path-specifications
 public struct XRPPathElement: Equatable {
+
+  /// If present, this path step represents rippling through the specified address.
+  /// This field is optional. MUST NOT be provided if this path element specifies the currency or issuer fields.
   public let account: Address?
+
+  /// If present, this path element represents changing currencies through an order book.
+  /// This field is optional. The currency specified indicates the new currency. MUST NOT be provided if this path
+  /// element specifies the account field.
   public let currency: XRPCurrency?
+
+  /// (Optional) If present, this path element represents changing currencies and this address defines the issuer of
+  /// the new currency. If omitted in a path element with a non-XRP currency, a previous element of the path defines the
+  /// issuer. If present when currency is omitted, indicates a path element that uses an order book between same-named
+  /// currencies with different issuers.
+  /// MUST be omitted if the currency is XRP. MUST NOT be provided if this element specifies the account field.
   public let issuer: Address?
 }
--- a/XpringKit/XRPPathElement.swift
+++ b/XpringKit/XRPPathElement.swift
@@ -2,12 +2,12 @@
 /// - SeeAlso: https://xrpl.org/paths.html#path-specifications
 public struct XRPPathElement: Equatable {
 
-  /// If present, this path step represents rippling through the specified address.
-  /// This field is optional. MUST NOT be provided if this path element specifies the currency or issuer fields.
+  /// (Optional) If present, this path step represents rippling through the specified address.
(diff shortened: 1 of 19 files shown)


austin-king:
> Documentation looks good, but looks like triple slash directives are not necessary for all of these comments?
> https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

I also have never used /// and just googled it though so I might just be misunderstanding their use case

hbergren:
@austin-king , this is Swift code, not TypeScript.

XpringKit/XRPPayment.swift
```diff
@@ -5,11 +5,24 @@ import Foundation
 public struct XRPPayment: Equatable {
   // TODO(keefertaylor): Modify this object to use X-Address format.
 
+  /// The amount of currency to deliver.
   public let amount: XRPCurrencyAmount
+
+  /// The unique address of the account receiving the payment.
   public let destination: Address
+
+  /// (Optional) Arbitrary tag that identifies the reason for the payment.
```
hbergren, line 14: Tags don't identify the reason for the payment, they indicate which user to send the payment to.

For example, Coinbase probably has one (or a handful) of actual XRP addresses, but each Coinbase user gets a tag number. So you send the XRP to Coinbase, and then Coinbase uses the tag number to credit the respective user inside their system.

XpringKit/XRPTransaction.swift
```diff
@@ -3,22 +3,56 @@ import Foundation
 /// A transaction on the XRP Ledger.
 ///
 /// - SeeAlso: https://xrpl.org/transaction-formats.html
-// TODO(keefertaylor): Modify this object to use X-Address format.
+/// TODO(keefertaylor): Modify this object to use X-Address format.
```
hbergren, line 6: Q: Are all comments in Swift `///`? Even TODOs?

amiecorso: Good question - I learned that no they're not:  multiline comments with `/** **/` and single line comments with `///` will show up in the generated interface, whereas multis with `/* */` and singles with just `//` won't.  This has been reverted.

XpringKit/XRPPayment.swift
amiecorso: I've updated this comment across all three SDKs... Check out the new version?  It's based on the documentation on xrpl.org docs here (under destinationTag): https://xrpl.org/payment.html.  Do you think that works, or should I entirely get rid of the bit about reasons?

austin-king:
> @austin-king , this is Swift code, not TypeScript.

hahaha I'm an idiot

XpringKit/XRPPayment.swift
hbergren: I think it's reasonable to use that language if that's what `xrpl.org` says.

XpringKit/XRPPathElement.swift
```diff
@@ -1,7 +1,20 @@
 /// A path on the XRPLedger
 /// - SeeAlso: https://xrpl.org/paths.html#path-specifications
 public struct XRPPathElement: Equatable {
+
+  /// If present, this path step represents rippling through the specified address.
```
keefertaylor, line 5: It's odd that You prefix these with 'this field is optional' and you use '(optional)' on the third? Any reason to not use consistent formatting?

amiecorso: Nope - I think that is a relic of grabbing comments directly from xrpl.org, in which some of them use an "(Optional)" at the beginning and others don't.  I updated a bunch of this in `xpring4j` while explicitly transforming fields to `Optional` types so I'll actually clean this up here real quick.

PR #197, opened 2020-04-16
Updates XRPTransaction class with additional data fields

amiecorso:
The `XRPTransaction` class is a native SDK class for wrapping protobuf objects.  This PR updates this class to maintain consistency with Xpring-JS by adding the fields `timestamp`, `hash`, and `deliveredAmount`.  This also requires that an `XRPTransaction` object be constructed from a `GetTransactionResponse` protobuf instead of just a `Transaction` protobuf, since these additional fields are found in the metadata of this response object.

Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/199

### Context of Change
In the process of migrating the Xpring Wallet to rely on the SDK for `paymentHistory` (in `transactionHistoryForWallet`) it was realized that the SDK's notion of a transaction (`XRPTransaction`) was missing some fields that the Wallet's notion of a transaction requires.  These changes were made in `Xpring-JS` and need to be propagated to the other two SDK flavors.

### Type of Change

## Before / After
`XRPTransaction` contains additional fields `hash`, `timestamp`, and `deliveredAmount`

## Test Plan
CI.  Tests have been updated with appropriate fake objects and test values.


amiecorso:
Keefer - as in the Java PR, I need help with the `toHex` function in `JavaScriptUtils`, which I should plumb through from `xpring-common-js` but am having trouble doing correctly.

XpringKit.xcodeproj/project.pbxproj
```diff
@@ -9,7 +9,6 @@
 /* Begin PBXBuildFile section */
 		000EF674D02FE63D72B1E1C3 /* AccountBalance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1585BA85CD7FF784C62F1885 /* AccountBalance.swift */; };
```
nkramer44, line 10: Not related to this PR at all, but should we keep the xcode project files in git?  I feel like they are super prone to bad merges and have to get regenerated anyway

XpringKit/Org_Xrpl_Rpc_V1_Transaction+XRPTransaction.swift
```diff
@@ -1,7 +1,13 @@
 import Foundation
 
 internal extension XRPTransaction {
-  init?(transaction: Org_Xrpl_Rpc_V1_Transaction) {
+  private static let javaScriptUtils = JavaScriptUtils()
+
+  init?(getTransactionResponse: Org_Xrpl_Rpc_V1_GetTransactionResponse) {
```
nkramer44, line 6: I think Keefer once told me that it's more idiomatic in Swift to do something like `init?(from: Org_Xrpl_Rpc_V1_Transaction).  Makes the parameter name more informative and cleans up the call site

Tests/DefaultXRPClientTest.swift
```diff
@@ -458,7 +458,7 @@ final class DefaultXRPClientTest: XCTestCase {
       .testTransactionHistoryResponse
       .transactions
       .map { transactionResponse in
-        return XRPTransaction(transaction: transactionResponse.transaction)
+        return XRPTransaction(getTransactionResponse: transactionResponse)
```
hbergren, line 461: Why did this change from `transaction` to `getTransactionResponse`?

Tests/ProtocolBufferConversionTests.swift
```diff
@@ -448,6 +448,9 @@ final class ProtocolBufferConversionTests: XCTestCase {
 
   func testConvertPaymentTransactionAllCommonFieldsSet() {
     // GIVEN a Transaction protocol buffer with all common fields set.
+    let hash = Data([2, 4, 6])
+    let timestamp: UInt32 = 0
+    let deliveredAmount: UInt64 = 20
```
hbergren, line 453: Are these the 3 fields that were added to Transactions in TransactionHistory?

XpringKit/JavaScript/JavaScriptUtils.swift
```diff
@@ -121,4 +121,11 @@ internal class JavaScriptUtils {
     }
     return result.toString()
   }
+
+  /// Convert the given byte array to a hexadecimal string.
+  /// - Parameter bytes: An array of bytes
+  /// - Returns: An encoded hexadecimal string.
+  public func toHex(bytes: Data) -> String {
+    return "STUB toHex"
```
hbergren, line 129: Is this to be implemented in the future? Or is it just for testing purposes?

XpringKit/XRPTransaction.swift
```diff
@@ -18,6 +19,8 @@ public struct XRPTransaction: Equatable {
   public let sourceTag: UInt32?
   public let transactionSignature: Data
   public let type: XRPTransactionType
+  public let timestamp: UInt64?
+  public let deliveredAmount: String?
```
hbergren, line 23: Just curious, when does timestamp get populated? When a transaction gets validated?

XpringKit.xcodeproj/project.pbxproj
amiecorso: @keefertaylor I have wondered this too.

XpringKit/Org_Xrpl_Rpc_V1_Transaction+XRPTransaction.swift
amiecorso: that may be true, though I wrote this to match the style of the other 8 wrapper classes that Keefer wrote :)  @keefertaylor can let me know if all of them should be updated with a `from` argument label

Tests/DefaultXRPClientTest.swift
amiecorso: a `GetTransactionResponse` contains a `Transaction`.  Originally, a `Transaction` had everything we needed to represent a transaction on the XRPL.  However, we realize that we actually want our notion of a transaction to include some metadata (such as the fields being added in this PR) which are only available in the surrounding object, a `GetTransactionResponse`.  Thus the change.

Tests/ProtocolBufferConversionTests.swift
amiecorso: I'm not sure I totally understand the question... but these 3 fields are fields that are present in the XpringWalletSystem's notion of a transaction.  Therefore we needed to plumb them through in the SDK in order for the SDK to be sufficient as an underlying messenger for `transactionHistoryForWallet` of the `XpringWalletSystemClient`.

XpringKit/JavaScript/JavaScriptUtils.swift
amiecorso: This is to be implemented in the future.  You'll see an explicit note to @keefertaylor in this PR because I need help wiring up the `xpring-common-js` module (in which `toHex` already resides) to be exposed here in the Swift wrapper class to that module.

XpringKit/XRPTransaction.swift
amiecorso: The answer to this is proving more elusive than I would have thought.  

On [this page](https://xrpl.org/transaction-basics.html#example-executed-transaction-with-metadata), in the example JSON response, we see a `date` field (represented as seconds after the Ripple epoch).  However, I haven't been able to find any discussion of this field in any of the other `xrpl.org` pages. 

According to a similar question I had for Elliot in the context of `ripple-lib`, it looks like transactions *sometimes* have a date, and if they don't, then `ripple-lib` uses the ledger's closing time as the timestamp for the transaction and returns that.  Elliot points out in a [slack archive](https://ripplelabs.slack.com/archives/C02GE9NBA/p1550776640039600?thread_ts=1550775985.037200&cid=C02GE9NBA) that lowercase fields are synthetic fields, whereas upper-case fields are generated directly from the server... `date` is a lower-case field, so it seems likely that transactions don't actually have a date on the XRP ledger, and that the ledger's close time is what's being subbed in for that.  Intuitively, this would make sense to me because a transaction isn't valid until it's included on a closed ledger and in that sense doesn't really exist in a meaningful way yet.

I'm asking Nik Bougalis.

amiecorso: Confirmed:  it's the close time of the ledger in which the transaction was included.

Tests/ProtocolBufferConversionTests.swift
hbergren: Are these fields not part of a `Transaction` object that you get from `rippled`?

amiecorso: These three fields are part of a `GetTransactionResponse` protobuf object that we get from `rippled` via `gRPC`

XpringKit/JavaScript/JavaScriptUtils.swift
keefertaylor: There's actually already functionality to do this in xpringkit:
https://github.com/xpring-eng/XpringKit/blob/master/XpringKit/XRP/ByteArray%2BHex.swift#L3

This is an extension on an array where the type of the array is a `uint8`. In english, anytime you have an array of bytes a special `toHex()` method gets added. 

`Data` is just a wrapper for a `[Uint8]` type in swift. So you can convert between the two fairly easily:
```
let data = Data()
let bytes = [UInt8](data)
let hex = byteArray.toHex()
```

Aside, I usually call `Data` types `data` and `[Uint8]` types bytes, though this is ultimately just a personal preference.

amiecorso: Awesome, also easy! This should be fixed and ready for review now.

Tests/ProtocolBufferConversionTests.swift
```diff
@@ -448,6 +448,9 @@ final class ProtocolBufferConversionTests: XCTestCase {
 
   func testConvertPaymentTransactionAllCommonFieldsSet() {
     // GIVEN a Transaction protocol buffer with all common fields set.
+    let hash = Data([2, 4, 6])
+    let timestamp: UInt32 = 0
```
keefertaylor, line 452: Similar to the comment on Java, I think this should be UInt64 since adding 946,684,800 + <arbitrary int value> could overflow. 

The (documentation also notes this](https://xrpl.org/basic-data-types.html#specifying-time):
> Don't convert Ripple Epoch times to UNIX Epoch times in 32-bit variables: this could lead to integer overflows.

amiecorso: So this particular value is just being fed to construct the `Date` protobuf... which requires a UInt32.  The `timestamp` field in the actual `XRPTransaction` object is indeed a `UInt64`, and the `UInt32` coming out of the protobuf gets converted to a `UInt64` before conversion, so I think we're good to go here.

PR #198, opened 2020-04-21
Subdirectory XRP component

amiecorso:
This PR reorganizes the XRP component files in `XpringKit` to reflect the organizational structure of both the other components and the other SDKs, including a subdirectory `model/` within `XRP/` to hold the XRP protobuf wrapper classes.

### Context of Change
Maintain organizational consistency across SDKs and components.

### Type of Change

## Test Plan
CI


amiecorso:
Keefer, maybe take a glance and let me know if any of these files actually belong somewhere better like `Common/`.  To my knowledge they were all created in the context of the `XRPClient` so I stuck them all in XRP.

PR #201, opened 2020-04-28
Validate Networks in XpringClient

keefertaylor:

Throw an error if XpringClient is initialized with objects that are bound to different networks.

### Context of Change

`XpringClient` composes functionality across components. It would not make sense to have components be operating on different networks as they try to interoperate. Provide a sanity check at construction time for this case.

Note that while the `network` property exists in `PayIDClient` and `XRPClient` they are not exposed in the interface and need to be created on `FakePayIDClient` and `FakeXRPClient`.

Introduce a `XpringError` to handle these cases. Throw a XpringError if this precondition is not met.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Safer code, fewer sharp edges.

## Test Plan

Additional unit test introduced to prove correctness.


Tests/Xpring/XpringClientTest.swift
```diff
@@ -154,4 +154,27 @@ final class XpringClientTest: XCTestCase {
 
     self.wait(for: [completionCalledExpectation], timeout: 10)
   }
+
+  func testMismatchedNetworks() throws {
+    // GIVEN a PayIDClient and an XRPClient on different networks.
+    let expectedTransactionHash = "deadbeefdeadbeefdeadbeef"
```
austin-king, line 160: lol what an odd string to use but ok

Tests/Xpring/XpringClientTest.swift
```diff
@@ -154,4 +154,27 @@ final class XpringClientTest: XCTestCase {
...
+      getBalanceValue: .success(fakeBalanceValue),
+      paymentStatusValue: .success(fakeTransactionStatusValue),
+      sendValue: .success(expectedTransactionHash),
+      latestValidatedLedgerValue: .success(fakeLastLedgerSequenceValue),
+      rawTransactionStatusValue: .success(fakeRawTransactionStatusValue),
+      paymentHistoryValue: .success(fakePaymentHistoryValue),
+      accountExistsValue: .success(fakeAccountExistsValue)
+    )
+
+    let fakeResolvedPayID = "r123"
```
austin-king, line 172: fakeResolvedPayID makes me thing it'll be PayID format, my impression is that this is supposed to be an XRP classic address (starts with r, random stuff after). I'd find it helpful if this variable was renamed

PR #203, opened 2020-05-03
Adds `xrpToDrops` and `dropsToXrp` to XRP/Utils

amiecorso:
This PR adds the utility functions `xrpToDrops` and `dropsToXrp` to `XrpingKit/XRP/Utils`.  These utility methods originally existed in `ripple-lib`, which is a robust and well-tested implementation that can handle either string or numeric values.  Their implementation here is a translation into Swift.  Note, however, that the Swift (and Java) methods currently only handle String arguments (whereas the JS implementation also handles BigNumber and Number types). Test cases are also based on the cases included in `ripple-lib` and added to `Tests/UtilsTest`.
 
The logic in the conversions begins with a regex to determine whether the argument is a well-formed number, and includes additional checks such as (in `dropsToXrp`) that any drops amount submitted as a decimal is actually a whole number (i.e. 5.00, 0.0 etc.) and (in `xrpToDrops`) that any xrp amount submitted as a decimal doesn't contain more than 6 non-zero numbers after the decimal, as the 7th and beyond would represent a fractional amount in drops - not allowed.

Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/399
Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/217

### Context of Change
Since this is such a common conversion, a robust implementation should be part of the SDKs. 

### Type of Change

## Before / After
`xrpToDrops` and `dropsToXrp` can now be imported from `XpringKit`

## Test Plan
New test cases pulled from `ripple-lib` to explore all conversion logic.


XpringKit/XRP/Utils.swift
```diff
@@ -64,4 +66,75 @@ public enum Utils {
...
+      .description(withLocale: Locale(identifier: "en_US"))
+  }
+
+  /// Convert from units in XRP to drops.
+  ///
+  /// - Parameter xrp: An amount of XRP expressed in units of XRP.
+  /// - Returns: A String representing an amount of XRP expressed in units of drops.
+  /// - Throws: An XRPLedgerError if XRP is in invalid format.
+  public static func xrpToDrops(_ xrp: String) throws -> String {
+    // TODO: nil preconditions check?
```
keefertaylor, line 110: you can resolve this. `String` is always non-nil in swift. If your parameter was a `String?` then you'd need to `guard let` here.

XpringKit/XRP/Utils.swift
```diff
@@ -64,4 +66,75 @@ public enum Utils {
...
+
+    let xrpRegex: String = "^-?[0-9]*['.']?[0-9]*$"
+    if xrp.range(of: xrpRegex, options: .regularExpression) == nil {
```
keefertaylor, line 113: I know it's arbitrary, but `guards` are more appropriate here. The language forces guard statements to return from the function inside of them which is a useful guard rail.

XpringKit/XRP/XRPLedgerError.swift
```diff
@@ -9,6 +9,12 @@ public enum XRPLedgerError: Equatable, Error {
   /// An invalid address was given. A more descriptive string is provided as an associated value.
   case invalidInputs(String)
 
+  /// An invalid drops amount was given.  A more descriptive string is provided as an associeted value.
```
keefertaylor, line 12: I'd bundle these into invalid inputs unless you really think they're needed.

amiecorso: nope.  was thrown off by the 'an invalid address was provided' comment but I just updated that to make it more generic.

PR #204, opened 2020-05-03
Enforce Ledger Specifiers as to not use Open Ledgers

amiecorso:
Tyler S noticed in [this PR](https://github.com/xpring-eng/Xpring-JS/pull/423) that currently, when requesting account information we are not setting the ledger specifier to validated ledgers only. This means all the balances and other account information we return can include unconfirmed values. This could lead to issues where a user thinks they have received a payment when in reality that payment could be fraudulent.

NOTE: 
Setting a `LedgerSpecifer` to `Shortcut.VALIDATED` in the `GetAccountTransactionHistoryRequest` was causing integration tests to fail.  After communicating with CJ Cobb about the rippled gRPC implementation I learned (paraphrased from CJ's responses):
- `Shortcut.VALIDATED` refers to the single most recently validated ledger held by a rippled node. 
- For account info, `Shortcut.VALIDATED` should be specified. Otherwise, it will use the current ledger (potentially open or closed but not validated), which would give back a balance (amongst other info) that is not yet validated.
- For account transaction history (the rippled JSON API equivalent is account_tx), that rpc *only ever returns validated data.*  You do not need to set a ledger specifier in this case, and doing so likely causes only a partial transaction history to be returned.
- To get info on a transaction that has not yet been validated, you have to call GetTransaction (the JSON equivalent is tx).  In this case, you can only specify an (optional) ledger range, and that is only used in error handling: if the transaction is present, that rpc returns the transaction, regardless of what you pass as the ledger range. If it's not present, the response will specify whether the entire range you specified was searched.  The idea is that you can know what ledgers the server did search, since servers can have gaps in the ledger history, which could lead to false negatives.
- Finally, the behavior of rippled gRPC calls is designed to mimic the JSON API as closely as possible, which has better documentation.

Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/423 (I will submit a revised PR for `xpring-js` that doesn't set a ledger specifier for full transaction history)
Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/218

### Type of Change

## Before / After
No more unconfirmed transactions impacting the user's balance or other account information.

## Test Plan
CI passes


XpringKit/XRP/DefaultXRPClient.swift
```diff
@@ -198,6 +201,11 @@ extension DefaultXRPClient: XRPClientDecorator {
 
     let request = Org_Xrpl_Rpc_V1_GetTransactionRequest.with {
       $0.hash = transactionHashData
+      $0.ledgerRange = Org_Xrpl_Rpc_V1_LedgerRange.with {
```
keefertaylor, line 204: I validated with Mark Travis on rippled that we don't ned to set this. If unset, all ledgers are searched. 

You can omit this code.

PR #206, opened 2020-05-04
Fixes Bug in X-Address Encoding

amiecorso:
Fix bug where encoding addresses without tags returned incorrect results.

Swift analogue to: https://github.com/xpring-eng/xpring4j/pull/221

### Context of Change
The signature of encodeXAddress in JavaScript is address: string, tag: number, isTest: bool. Without a tag, we were passing address: string, isTest: bool which was yielding bad results.

### Type of Change

## Before / After
Added a new test case and ensured it failed. Corrected code and ensured it passed.

## Test Plan
Circle CI


XpringKit/JavaScript/JavaScriptUtils.swift
```diff
@@ -80,6 +80,9 @@ internal class JavaScriptUtils {
     var arguments: [Any] = [ classicAddress ]
     if tag != nil {
       arguments.append(tag as Any)
+    } else {
+      let undefined = JavaScriptLoader.load("undefined", from: JSValue(undefinedIn: <#T##JSContext!#>))
```
keefertaylor, line 84: I think that if you just pass nil it will evaluate to `undefined` when converted to JS. See table 1 here: https://developer.apple.com/documentation/javascriptcore/jsvalue

amiecorso: Note: Keefer and I discovered that this apple documentation is wrong, and `nil` was actually converting to `0` in JS.  We instead use `JSValue(undefinedIn: JSContext)` to represent an absent argument.

PR #216, opened 2020-05-15
Fix semantics of methods for current open ledger and last validated ledger

keefertaylor:

The current implementation of `getLatestValidatedLedgerSequence` returns the latest *open* ledger sequence. 

Fix this method's name and re-implement `getLatestValidatedLedgerSequence`.

### Context of Change

There are a number of assumptions required to get `getLatestValidatedLedgerSequence` working. Namely, an address that is guaranteed to exist needs to be provided. 

I've worked with @carlhua  and @cjcobb23  on the rippled team and we are convinced that this implementation will never fail to return the latest validated ledger sequence when called from our reliable submission logic. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Reliable submission bug is fixed. 

## Test Plan

CI

### DIFF
--- a/Tests/Fakes/FakeXRPClient.swift
+++ b/Tests/Fakes/FakeXRPClient.swift
@@ -1,4 +1,4 @@
-import XpringKit
+@testable import XpringKit
 
 /// A  fake XRPClient which returns the given iVars as results from XRPClientDecorator calls.
 ///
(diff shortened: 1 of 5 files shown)


keefertaylor:
> Very nice. Question: is there any reason for parameterizing getLatestValidatedLedgerSequence with an address, when you could hard-code the genesis address instead? it seems less likely than any other address to have been deleted or somehow not exist (though I suppose it puts some responsibility on us v.s. the user in terms of picking an extant address..)

The genesis address can be deleted, because someone rotated the keys on the account and they have control of the account. 

On the other hand, we're sure that the address which sent a transaction will exist for the duration of the time that the transaction is in flight (unless the account is deleted in a race condition, in which case the transaction will fail anyway). 

> Then there's just the piece we discussed offline about potentially returning (hash, status) tuples from this method instead of just a hash that doesn't tell you anything about the actual ultimate status of the transaction, but that would be a separate PR anyway

Yup, trying to keep this to be a minimal change.

PR #220, opened 2020-05-16
Adds `validated` and `ledgerIndex` fields to `XRPTransaction`

amiecorso:
This PR adds the `validated` and `ledgerIndex` fields to the `XRPTransaction` object. 
These fields come from a `GetTransactionResponse` protobuf object, which is returned by a call to `getTransaction`. 

### Context of Change
This change is made in preparation for implementing a `getTransaction` method in the SDKs. The original implementation of the `XRPTransaction` class in the SDK only included the fields present in the `Transaction` protocol buffer object (which is wrapped by a `GetTransactionResponse` also containing other metadata about the transaction's outcome.) Certain of these other metadata fields have been added as-needed to the `XRPTransaction` object, and this PR takes care of the remaining fields available in `GetTransactionResponse`.

Our current approach is to combine the distinct `GetTransactionResponse` and `Transaction` protocol buffer objects into one type (`XRPTransaction`) since the metadata about the transaction's outcome is of equal importance to the specific transaction data, and while the type distinction makes sense in the context of authoring rippled, it would add unnecessary complexity in the SDK. 

### Type of Change

## Before / After
Two additional fields available on an `XRPTransaction` object.

## Test Plan
Added new test values to fake protobuf objects for conversion tests.
CI passes.

### DIFF
--- a/Tests/ProtocolBufferConversionTests.swift
+++ b/Tests/ProtocolBufferConversionTests.swift
@@ -462,6 +462,9 @@ final class ProtocolBufferConversionTests: XCTestCase {
     let memoData = Data([1, 2, 3])
     let memoFormat = Data([4, 5, 6])
     let memoType = Data([7, 8, 9])
+    let validated = true
+    let ledgerIndex = UInt32(1_000)
+
     let memoProto = Org_Xrpl_Rpc_V1_Memo.with {
       $0.memoData = Org_Xrpl_Rpc_V1_MemoData.with {
         $0.value = memoData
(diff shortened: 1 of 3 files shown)


Tests/ProtocolBufferConversionTests.swift
```diff
@@ -462,6 +462,9 @@ final class ProtocolBufferConversionTests: XCTestCase {
     let memoData = Data([1, 2, 3])
     let memoFormat = Data([4, 5, 6])
     let memoType = Data([7, 8, 9])
+    let validated = true
+    let ledgerIndex = UInt32(1_000)
```
tedkalaw, line 466: can these be consts? they don't seem to be mutated at all

amiecorso: let is the Swift version of const :)

PR #222, opened 2020-05-20
Implements `getPayment` in `XRPClient`

amiecorso:
This PR implements `getTransaction` in `XRPClient` and all related classes, which queries the XRP ledger for a transaction by hash, and if found, returns an `XRPTransaction` object representing the transaction.  Unit and integration tests added.

Note: I also moved some of the test files into the `XRP` and `XRP/Fakes` subdirectories to stay organized, and had to add a few test objects to the extensions in `Helpers` because we haven't done a refactor to use all global test objects in this SDK yet.

Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/230
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/470

### Context of Change
Mayur mentioned that a partner (presumably Forte) was looking for this functionality in the SDK. 

### Type of Change

## Before / After
`XRPClient` now exposes `getTransaction`

## Test Plan
New unit and integration tests pass on CI. 

## Future Tasks

This is the third and final SDK to be updated.


Tests/XRP/Fakes/FakeXRPClient.swift
```diff
@@ -71,6 +75,15 @@ extension FakeXRPClient: XRPClientDecorator {
     return try returnOrThrow(result: accountExistsValue)
   }
 
+  public func getPayment(for transactionHash: TransactionHash) throws -> XRPTransaction? {
+    switch getPaymentValue {
```
keefertaylor, line 79: You can just use the `returnOrThrow` helper method which does the same thing as your logic.

hbergren:
I'll review once tests pass in case functionality needs to change.

keefertaylor:
Sunday musing: Should we update the README.md file with a tutorial for this as well?

XpringKit.podspec
```diff
@@ -14,7 +14,7 @@ Pod::Spec.new do |spec|
   spec.author       = { "Xpring Engineering" => "xpring@ripple.com" }
   spec.source       = { :git => "https://github.com/xpring-eng/XpringKit.git", :tag => "v3.1.1" }
 
-  spec.swift_versions = [4.2]
+  spec.swift_versions = [5.1]
```
hbergren, line 17: what does this bump mean?

amiecorso: Ah yes, so the latest version of Swift is actually 5.2.  (Keefer suggested 5.1 which I imagine might just be caution around the cutting edge)  I was getting warnings-as-errors during the CI build for some code in this PR because the podspec version was out of date while my local version of Swift was up to date, so sneaking that version bump in here atomically to get the code to pass.

Tests/XRP/Fakes/FakeXRPClient.swift
amiecorso: so the reason I had to roll my own here is that `returnOrThrow` requires the type `Result<T, XRPLedgerError>`, but I needed to switch on type `Result<XRPTransaction?, RPCError>`

PR #223, opened 2020-05-22
Updates in response to breaking PayID changes - Public API

amiecorso:
This PR updates the SDK to handle breaking changes made to PayID.  Specifically, the `PaymentInformation` type was changed to return addresses in the form of an array of `Address` (new type) objects. 

Apologies for the 23-file change - this is intentional because I had to change the name of `Address.swift` (which was simply a type alias for String) to `XRPAddress.swift` (and all instances throughout the code) to avoid a name collision with the new PayID type `Address`. 
There are also 3 swagger-generated files present (since we have to check in the generated code).
Only two files were manually updated WRT breaking PayID changes:
- `pay-id.v1.yml`
- `PayIDClient.swift`

(The `XRPPayIDClient` unit tests avoid mocked networking in this SDK and therefore, unlike the other two SDKs, didn't need to be updated.)

analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/474
analogue in Java: https://github.com/xpring-eng/xpring4j/pull/245

@keefertaylor check out my TODO in `PayIDClient`.

----------------------------------------------------
New object structure in payid:

Screen Shot 2020-05-21 at 6 42 45 PM

### Type of Change

## Before / After
`PayIDClient` unit and integration tests pass again.

## Test Plan
CI


keefertaylor:
Brutal that we have to rename Address to XRPAddress. Can we note that in the CHANGELOG?

hbergren:
> Brutal that we have to rename Address to XRPAddress. Can we note that in the CHANGELOG?

Would it be easier to make `Address => PayIdAddress` instead?

XpringKit/PayID/Generated/Models/Address.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    public init(paymentNetwork: String, environment: String?, addressDetailsType: String, addressDetails: CryptoAddressDetails) {
+        self.paymentNetwork = paymentNetwork
+        self.environment = environment
+        self.addressDetailsType = addressDetailsType
+        self.addressDetails = addressDetails
+    }
+
+
+}
```
hbergren, line 27: I remember asking this before, but I forgot the answer. Why do we check in generated code again?

pay-id-api-spec/pay-id.v1.yml
```diff
@@ -148,21 +148,38 @@ components:
           type: string
       required:
         - address
-    PaymentInformation:
-      title: PaymentInformation
+    Address:
```
hbergren, line 151: Can we make this "PayIdAddress" in the SDK swagger? Then you might not need to rename `XRPAddress`, right?

XpringKit/PayID/PayIDClient.swift
```diff
@@ -71,7 +71,8 @@ public class PayIDClient {
       if
         let response = response,
         let paymentInformation = response.body {
-        completion(.success(paymentInformation.addressDetails))
+        // TODO(amiecorso): what if addresses is empty or has more than one address?
+        completion(.success(paymentInformation.addresses[0].addressDetails))
```
hbergren, line 75: Great point. The whole reason for the breaking change was to allow requesting `Accept: application/payid+json`, which returns the array of all addresses that PayID has. Does the SDK not support that right now?

amiecorso:
> > Brutal that we have to rename Address to XRPAddress. Can we note that in the CHANGELOG?
> 
> Would it be easier to make `Address => PayIdAddress` instead?

I think the issue with that is the fact that `Address` is an object that's part of the actual PayID protocol... so I feel like renaming `Address` to `PayIdAddress` actually has deeper implications than renaming `Address` to `XRPAddress`.

pay-id-api-spec/pay-id.v1.yml
amiecorso: yeah this is a good point... comments below, and want to ask Keefer before changing this.

XpringKit/PayID/Generated/Models/Address.swift
amiecorso: also one for @keefertaylor

XpringKit/PayID/PayIDClient.swift
amiecorso: ah gotcha.  No the SDK does not support this right now in the sense that it is not set up to return an array of addresses.  If a `PayIDClient` was instantiated with the network `"payid"` as the argument to the constructor, then it would assemble and submit the content negotiation accept header.  I think we need to decide how to handle this header:

XpringKit/PayID/Generated/Models/Address.swift
keefertaylor: Certain build systems in Swift (Carthage, Swift PM) pull in dependencies by checking out GitHub projects. These systems have hooks in them that run beforehand (and where we could generate this code) but the average user ended up having trouble with this because they didn't have the suite of tools installed they needed: grpc, node, git, protoc, and swagger. 

Instead, they'd get some random compile errors, think the project was broken because they didn't read the build logs and would open issues against the project. It's easier to just check them in than to force everyone who wants to use this project to install all the needed tooling.

XpringKit/PayID/PayIDClient.swift
keefertaylor: It is tough for me to imagine why we'd expose something that let you get multiple addresses in the SDK. The binding here is 1:1 between PayIDClient and Network. (It's tough for me to visualize in general why anyone would want content negotiation, but that's out of band). You'd never ask a PayIDClient connected to "xrpl-test" for a "eth" address. Simply put, this method should never return more than 1 address and it is an error condition if the network request does. 

The way I'd do PayID negotiation on the client would probably be composing an array of PayIDClients into a `MultiPayIDClient` (Name TBD). You can also come up with a nicer API where you give us an array of strings and we make you the PayIDClients. They, you just need to iterate through each `PayIDClient` in `MultiPayIDClient` and return the first one that gives you an address. That certainly seems easier to me than crafting some bespoke headers and trying to interpret some non-standardized string array sent back from the server. This isn't hard to do but no one has asked for it yet and I'd want to hear a use case before we expand the PayID footprint further.  

The only case I can imagine is where you'd want to retrieve all addresses for a user. I'm not really sure where that lives... It doesn't make sense in the context of a `PayIDClient` and it doesn't make sense in a context of a `MultiPayIDClient`. That's probably a third class? Or a utility method? TBD, but it also wouldn't live in this 1:1 hierarchy. 

----

I'm supportive of PayID adding content-negotiation if that's what the team, or the powers that be, want. There are way easier ways to do this client side than content negotiation.

pay-id-api-spec/pay-id.v1.yml
keefertaylor: This isn't really an 'Address' It's more of "AddressAndMetadata"? I actually don't have a great name here: "AddressResult"?

keefertaylor:
Meta: @amiecorso  if we end up changing the yaml, I would like to do a final round of deduplication on all the yamls we have after launch, so no need to worry about de-duping them across repos now.

pay-id-api-spec/pay-id.v1.yml
hbergren: Yeah, the actual "address" here is the `addressDetails`, right?

I'm ok with any name you guys think is appropriate. My vote would still be for `PayIdAddress`, which can encapsulate the concept that it is both metadata around a payment rail address, and the payment rail address itself.

XpringKit/PayID/PayIDClient.swift
hbergren: > It is tough for me to imagine why we'd expose something that let you get multiple addresses in the SDK.

I guess the question is whether or not we want the SDK to be the "normal" way to interact with PayID. I don't know this for certain, but I suspect that most partners are rolling their own HTTP APIs rather than using the SDK right now.

I have two (theoretical) concrete scenarios that illustrate content negotiation and "get all addresses".

#### Content Negotiation

I have my wallet with BTC/ETH/XRP. We go out for dinner together, and you cover the bill to get the credit card points, and I want to pay you back. Neither of us care whether I pay you in BTC, ETH, or XRP, and I may not even know which currencies your wallet actually supports.

The UI for this wallet might even just be a "PAY" button with a input for a PayID.

I want to pay the least amount in fees when transferring money to you, so I make a payment with `Accept: xrpl, eth;q=0.9, btc;q=0.8`. I only need to make a single HTTP request instead of multiple, and I'm guaranteed to get the payment address that will cost me the least in fees (assuming XRPL < ETH < BTC for fees).

#### Get All Addresses

My wallet has a UI allowing me to select which currencies I can pay you in. However, it also wants to disable currencies that the beneficiary doesn't support. I make my `payid+json` request to you, find out you don't support XRPL, and disable the XRP selection on the UI.

PR #227, opened 2020-05-29
Use Alternative Swagger Code Generator

keefertaylor:

Use an alternative Swagger generator that generates nicer / idiomatic code. 

### Context of Change

The built in Swift generator for Swagger is untestable, and it also declares the `host` variable staticly, meaning that there can only ever be one host at a time. 

This code appears to be testable and fixes the host URL issue. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change - internal refactor

## Test Plan

CI

## Future Tasks

PayID should now be unit testable. Will add tests in a follow up.


tedkalaw:
good

keefertaylor:
I've folded in a few changes here that are required to fix the build. A number of things colluded against us here causing this PR to be multi-purpose:
- Something with ruby in the CircleCI environment changed
- PayID pushed a new release

To fix:
- Update `PayIDClient` to include the `PayID-Version` header - fixes PayID breakage
- Update CircleCI config, roughly following instructions here: https://circleci.com/docs/2.0/testing-ios/#using-ruby

Not sure *what* broke, but it has something to do with ruby versioning. The fix:
- Update to the latest MacOS image on CircleCI (hunch that they broke the older image somehow) 
- Update the ruby command to the syntax needed by the latest image
- Manually install the `xcpretty` gem which is missing from the latest image. 
- Update the simulator type / version to match simulators available on the newer image

XpringKit/PayID/Generated/APIClient.swift
```diff
@@ -0,0 +1,252 @@
...
+        self.defaultHeaders = defaultHeaders
+        jsonDecoder.dateDecodingStrategy = .custom(dateDecoder)
+        jsonEncoder.dateEncodingStrategy = .formatted(API.dateEncodingFormatter)
+    }
+
+    /// Makes a network request
+    ///
+    /// - Parameters:
+    ///   - request: The API request to make
+    ///   - behaviours: A list of behaviours that will be run for this request. Merged with APIClient.behaviours
```
LoisRP, line 44: behaviours is a British/Commonwealth spelling. It's not wrong per se, but you might want to reconsider.

LoisRP: It's probably not worth changing at this point.

keefertaylor: This is generated code, so we're at the mercy of the British engineer who forced this upon us.

PR #233, opened 2020-05-30
Adds X-Addresses to `XRPTransaction` and `XRPPayment` model objects

amiecorso:
This PR adds X-address to the `XRPPayment` and `XRPTransaction` model objects.  `XRPPayment` contains a `destination` and a `destinationTag` property, and now also contains a `destinationXAddress` property that is the X-address encoded version of this information. `XRPTransaction` contains an `account` property, and now also contains an `accountXAddress` property, which is the X-address encoded version of the originating account with no destination tag.

In order to create the correct X-address encoding, we need to know whether we're on Testnet or Mainnet.  Therefore the constructors for `XRPPayment` and `XRPTransaction` now have an additional `xrplNetwork: XRPLNetwork` parameter.  Tests and client code have been updated accordingly (including addition of this network param to the `XRPClientDecorator` and implementors, since this needs to be plumbed through in order to construct accurate `XRPTransaction`s)

Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/487
Analogue in Java: https://github.com/xpring-eng/xpring4j/pull/250

### Type of Change

## Before / After
`XRPTransaction` and `XRPPayment` now contain X-address versions of address and tag properties. 

## Test Plan
Unit tests for protocol buffer conversion into these objects have been updated to verify the X-address properties.
CI still passes.

## Future Tasks

Questions for @keefertaylor 
- do we need to deprecate the classic address/tag fields, or just leave everything side-by-side?
- should a note about this go in the CHANGELOG?  The `XRPTransaction` and `XRPPayment` objects have an additional constructor parameter, but it's optional (defaults to Mainnet) and thus isn't a breaking change.
- should other classes that contain XRP addresses (such as `XRPPathElement` or `XRPIssuedCurrency`) also contain X-address versions of these?

### DIFF
--- a/Tests/DefaultXRPClientTest.swift
+++ b/Tests/DefaultXRPClientTest.swift
@@ -11,7 +11,7 @@ final class DefaultXRPClientTest: XCTestCase {
   // MARK: - Balance
   func testGetBalanceWithSuccess() {
     // GIVEN an XRPClient which will successfully return a balance from a mocked network call.
-    let xrpClient = DefaultXRPClient(networkClient: FakeNetworkClient.successfulFakeNetworkClient)
+    let xrpClient = DefaultXRPClient(networkClient: FakeNetworkClient.successfulFakeNetworkClient, xrplNetwork: XRPLNetwork.test)
 
     // WHEN the balance is requested.
     guard let balance = try? xrpClient.getBalance(for: .testAddress) else {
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -205,6 +205,22 @@ let payment = try! xrpClient.getPayment(for: transactionHash)
...
+import xpringkit
+
+let remoteURL = "alpha.test.xrp.xpring.io:50051"; // TestNet URL, use alpha.xrp.xpring.io:50051 for Mainnet
+let xrpClient = XRPClient(grpcURL: remoteURL, useNewProtocolBuffers: true)
+
+let transactionHash = "9FC7D277C1C8ED9CE133CC17AEA9978E71FC644CE6F5F0C8E26F1C635D97AF4A"
+let payment = try! xrpClient.getPayment(for: transactionHash)
+```
+
+**Note:** The example transactionHash may lead to a "Transaction not found." error because the TestNet is regularly reset, or the accessed node may only maintain one month of history.  Recent transaction hashes can be found in the [XRP Ledger Explorer ](https://livenet.xrpl.org/).
```
LoisRP, line 222: Code format for transactionHash ?

PR #238, opened 2020-06-05
Initial Test for `PayIDClient`

keefertaylor:

Add testing infrastructure for PayID.

### Context of Change

Previous versions of Swagger were untestable. #227 migrated to a new generator which has testable code. 

This PR does a few things:
- Injects `SessionManager` into `PayIDClient` in the initializer
- Creates a `FakeURLProtocol` so that we can fake networking
- Hooks up a basic test

The faking code is heavily adapted from this blog post: https://www.google.com/search?q=alamofire+fake&oq=alamofire+fake&aqs=chrome..69i57.2366j0j1&sourceid=chrome&ie=UTF-8. 

Since it is mostly boiler plate and it's only test code, I didn't do much in the way of documentation or stylistic updating. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Better test coverage. 

## Test Plan

New test passes on CI. 

## Future Tasks
We should continue building tests for PayID. They are omitted here for brevity of this PR. Once the testing infrastructure exists it is trivial to write additional test cases.


Tests/PayID/PayIDClientTest.swift
```diff
@@ -0,0 +1,58 @@
...
+final class PayIDClientTest: XCTestCase {
+  func testAddressWithSuccessResponse() {
+    // GIVEN A PayIDClient with faked networking to return a successful response.
+    let address = "X7cBcY4bdTTzk3LHmrKAK6GyrirkXfLHGFxzke5zTmYMfw4"
+    let paymentNetwork = "xrpl-testnet"
+    let paymentInformation = PaymentInformation(
+      addresses: [
+        PayIdAddress(
+          paymentNetwork: paymentNetwork,
+          addressDetailsType: "",
```
dino-rodriguez, line 15: Did you intend to leave `addressDetailsType` as blank in this response?

Tests/Common/Helpers/FakeURLProtocol.swift
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  override func stopLoading() {
+    activeTask?.cancel()
+  }
+}
+
+// MARK: - URLSessionDataDelegate
+
+extension FakeURLProtocol: URLSessionDataDelegate {
```
dino-rodriguez, line 46: What's the reasoning for using the class extensions?

Tests/Common/Helpers/FakeURLProtocol.swift
```diff
@@ -0,0 +1,82 @@
...
+  }
+  static var responseType: ResponseType!
+
+  private lazy var session: URLSession = {
+    let configuration: URLSessionConfiguration = URLSessionConfiguration.ephemeral
+    return URLSession(configuration: configuration, delegate: self, delegateQueue: nil)
+  }()
+  private(set) var activeTask: URLSessionTask?
+
+  override class func canInit(with request: URLRequest) -> Bool {
```
dino-rodriguez, line 22: Some comments on these methods would be helpful to understand context on what these are for and how they are used together.

Tests/PayID/PayIDClientTest.swift
```diff
@@ -0,0 +1,58 @@
...
+    payIDClient.address(for: "georgewashington$xpring.money") { result in
+      // THEN the response contains the expected address.
+      switch result {
+      case .success(let addressDetails):
+        XCTAssertEqual(addressDetails.address, address)
+      case .failure(let error):
+        XCTFail("flagrant error: \(error)")
+      }
+
+      expectation.fulfill()
```
hbergren, line 51: If the network request is faked, then what exactly is being tested here? That creating a PayIDClient with a `xrpl-testnet` network doesn't throw an error?

XpringKit/PayID/PayIDClient.swift
```diff
@@ -29,8 +36,9 @@ public class PayIDClient {
   ///   - ach
   ///
   //  TODO: Link a canonical list at payid.org when available.
-  public init(network: String) {
+  public init(network: String, sessionManager: SessionManager = .default) {
```
hbergren, line 39: Is this a parameter now so we can fake network requests?

Tests/Common/Helpers/FakeURLProtocol.swift
keefertaylor: Generally when you conform to a protocol (`FakeURLProtocol` is conforming to `URLSessionDataDelegate`) you separate the functionality into an extension so it is clear what functions belong to what thing. 

Functionally, it is equivalent to just doing this at the class level.

XpringKit/PayID/PayIDClient.swift
keefertaylor: Correct. This is a manager for `AlamoFire`, which is the networking library that OpenAPI uses.

TBH, I don't know a ton about how this actually works, so I'm mostly just following tutorials on how to mock swagger code.

Tests/PayID/PayIDClientTest.swift
keefertaylor: The parsing logic and error handling logic is also tested. 

Admittedly, this is a pretty thin test. In the future we will have cases for:
- Malformed response
- Not found
- Generic network error

Tests/Common/Helpers/FakeURLProtocol.swift
keefertaylor: I added docs on the stuff that is unique. 

A lot of this is boilerplate implementations of protocols, which already have the doc. For instance, there's official documentation on [URLProtocol](https://developer.apple.com/documentation/foundation/urlprotocol) and the [`canInit(with:)` method](https://developer.apple.com/documentation/foundation/urlprotocol/1411389-caninit).

I attached `SeeAlso` tags to link to this documentation. 

LMK if you want more detail. I'm being somewhat vague because I don't really know the way all of these methods wires together and I'm really just following the linked tutorial.

Tests/PayID/PayIDClientTest.swift
hbergren: > The parsing logic and error handling logic is also tested.
>
> Admittedly, this is a pretty thin test. In the future we will have cases for:
>
> Malformed response
> Not found
> Generic network error

Got it. Is there a reason to prefer mocking network requests to test this stuff vs writing unit tests that cover the same areas? Is it that you want the E2E nature of a network test?

keefertaylor: In this case, I prefer mocking because I'm relying heavily on Swagger and not really using it as intended. I could certainly find a way to inject a swagger API client and verify that it works. 

This was originally a holdover from when we started doing just plain network requests (before OpenAPI code generation). Basically, the reason we moved to code generation was because of the schemas in PayID+ being more difficult. 

I don't really have an opinion here. I can always find a way to fake these things; in this particular instance I think mocking might give us a little more flexibility if we ever change the class implementation. I also see the argument that we should be consistent in our test strategy though. 

Overall, I'm motivated to try to land this as-is rather than re-writing as a fake but you can LMK if you feel strongly?

hbergren: No, I think network mocking makes sense here, if the other option is faking.

I'm just intuitively skeptical of mocks/fakes in tests, because then you need to think a little harder about what you're actually testing, but you've clearly thought that through, so it's fine.

PR #263, opened 2020-08-03
(3/3 swift) Implements `enableDepositAuth` on `XRPClient`

amiecorso:
This PR adds the method `enableDepositAuth` to the `XRPClient`, which enables [Deposit Authorization](https://xrpl.org/depositauth.html) on the specified account.  This is achieved by submitting an `AccountSet` transaction with a set flag set to 9 (asfDepositAuth) (see https://xrpl.org/accountset.html#accountset-flags).  An API for submitting generic `AccountSet` transactions is not yet exposed - this exposes only a specific piece of functionality.

Analogue in `xpring-js` : https://github.com/xpring-eng/Xpring-JS/pull/585
Analogue in `xpring4j` : https://github.com/xpring-eng/xpring4j/pull/288

### Context of Change

The Charity App needs to be able to enable deposit authorization, and we would like it to rely on the SDK. This is also serving as a preliminary attempt at reasoning about inclusion of new transaction types in a user-friendly way.

### Type of Change

- [x] Documentation Updates

## Before / After
`enableDepositAuth` exposed on `XRPClient`

## Test Plan
CI still passes.  New tests written for `enableDepositAuth`


Tests/Xpring/XpringClientTest.swift
```diff
@@ -41,7 +46,8 @@ final class XpringClientTest: XCTestCase {
       rawTransactionStatusValue: .success(fakeRawTransactionStatusValue),
       paymentHistoryValue: .success(fakePaymentHistoryValue),
       accountExistsValue: .success(fakeAccountExistsValue),
-      getPaymentValue: .success(fakeGetPaymentValue)
+      getPaymentValue: .success(fakeGetPaymentValue),
```
keefertaylor, line 49: Drive by observation: in the future we can maybe make these parameters optional or have a default?

XpringKit/XRP/Model/TransactionResult.swift
```diff
@@ -0,0 +1,19 @@
...
+  /// The identifying hash of the transaction.
+  public let hash: TransactionHash
+
+  /// The TransactionStatus indicating the outcome of this transaction.
+  public let status: TransactionStatus
+
+  /// Whether this transaction is included in a validated ledger.  The transactions status is only final if this field is true.
+  public let validated: Bool
+
+  /// Create a new TransactionResult object.
```
keefertaylor, line 13: You can actually omit this:
https://docs.swift.org/swift-book/LanguageGuide/Initialization.html#ID213

Swift will generate you initializers automatically if you use a struct and all the fields of the object are also structs or primitives.

amiecorso: removed

Tests/Xpring/XpringClientTest.swift
amiecorso: I agree - I was wondering how to handle this as it potentially grows larger and larger.  maybe can refactor all at once in future

PR #266, opened 2020-08-06
Adds Support for Transaction Memos in `XRPClient`

amiecorso:
Adds support for transaction memos via additional public API `sendWithDetails` exposed on `XRPClient` and `XpringClient`.  `sendWithDetails` takes a new object type, a `SendXrpDetails` that bundles the amount, destination, sending wallet, and list of memos.

Analogue in `xpring-js`: https://github.com/xpring-eng/Xpring-JS/pull/501
Analogue in `xpring4j`: https://github.com/xpring-eng/xpring4j/pull/299

### Context of Change
The initial implementation of `send` in `XRPClient` was kept intentionally simplistic, however, some business logic (such as ChefXRP) requires the use of the `Memos` option on transactions.  This PR enables that information to be optionally included in a transaction, while also leaving the original simplified API intact.

### Type of Change

- [x] Documentation Updates

## Before / After
Before: memos weren't addable to the transaction
After: memos can be added and created with hex and non-hex strings

## Test Plan
Added unit and integration tests for the new functionality, CI still passes.

### DIFF
--- a/XpringKit/XRP/Model/SendXRPDetails.swift
+++ b/XpringKit/XRP/Model/SendXRPDetails.swift
@@ -13,4 +13,16 @@ public struct SendXRPDetails {
 
   /// A list of memos to attach to the payment transaction constructed from these details.
   public let memosList: [XRPMemo]?
+  
+  public init(
+    _ amount: UInt64,
+    _ destination: Address,
+    _ sender: Wallet,
+    _ memosList: [XRPMemo]?
+  ) {
+    self.amount = amount
+    self.destination = destination
+    self.sender = sender
+    self.memosList = memosList
+  }
 }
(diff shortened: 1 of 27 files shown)


XpringKit/XRP/Model/SendXRPDetails.swift
```diff
@@ -0,0 +1,28 @@
+/// Describes the fine grained details for sending XRP. The destination field may be a PayID, XAddress,
```
tedkalaw, line 1: hella nit: `fine-grained`

XpringKit/XRP/Model/SendXRPDetails.swift
```diff
@@ -0,0 +1,28 @@
...
+  /// The receiving address.
+  public let destination: Address
+
+  /// The sending wallet.
+  public let sender: Wallet
+
+  /// A list of memos to attach to the payment transaction constructed from these details.
+  public let memosList: [XRPMemo]?
+
+  public init(
```
keefertaylor, line 17: you can omit this constructor and let Swift generate it for you.

amiecorso: I can't in this case actually because I need `SendXRPDetails` to have a publicly accessible constructor, and according to this I have to implement that myself: https://stackoverflow.com/questions/26224693/how-can-i-make-the-memberwise-initialiser-public-by-default-for-structs-in-swi

PR #273, opened 2020-08-19
(4/17) Creates `XRPCheckCash` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPCheckCash` - a native Swift struct to wrap the [`CheckCash` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L132).  See also: https://xrpl.org/checkcash.html

In this PR, I also took the liberty of creating some stand-alone test objects via extensions for common types like `Amount` which contains an `IssuedCurrencyAmount` etc.  
We've been meaning to refactor tests to use re-usable test objects, and this is a step in that direction as-needed to avoid letting things get messier.

### Context of Change

### Type of Change

## Before / After
`XRPCheckCash` class and conversion unit tests now exist.

## Test Plan


XpringKit/XRP/Model/XRPCheckCash+Org_Xrpl_Rpc_V1_CheckCash.swift
```diff
@@ -0,0 +1,34 @@
...
+  ///     - checkCash: an Org_Xrpl_Rpc_V1_CheckCash (protobuf object) whose field values will be used to
+  ///                  construct an XRPCheckCash
+  /// - Returns: an XRPCheckCash with its fields set via the analogous protobuf fields.
+  init?(checkCash: Org_Xrpl_Rpc_V1_CheckCash) {
+    // checkId is required
+    self.checkId = String(decoding: checkCash.checkID.value, as: UTF8.self)
+    if self.checkId.isEmpty {
+      return nil
+    }
+    // amount and deliverMin fields should be mutually exclusive
```
amiecorso, line 22:
```suggestion

    // amount and deliverMin fields should be mutually exclusive
```

PR #274, opened 2020-08-19
(5/17) Creates `XRPCheckCreate` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPCheckCreate` - a native Swift struct to wrap the [`CheckCreate` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L145).  See also https://xrpl.org/checkcreate.html

### Context of Change

### Type of Change

## Before / After
`XRPCheckCreate` class and conversion unit tests now exist.

## Test Plan


XpringKit/XRP/Model/XRPCheckCreate+Org_Xrpl_Rpc_V1_CheckCreate.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+    self.destinationXAddress = destinationXAddress
+
+    // sendMax is required
+    let sendMaxCurrencyAmount = checkCreate.sendMax.value
+    guard let sendMax = XRPCurrencyAmount(currencyAmount: sendMaxCurrencyAmount) else {
+      return nil
+    }
+    self.sendMax = sendMax
+
```
amiecorso, line 36: For the below fields (expiration and invoiceId) - is this the cleanest and most idiomatic way to achieve what we're doing here?

keefertaylor: I think you could use optional chaining: https://docs.swift.org/swift-book/LanguageGuide/OptionalChaining.html

PR #276, opened 2020-08-20
(7/17) Creates `XRPEscrowCancel` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPEscrowCancel` - a native Swift struct to wrap the [`EscrowCancel` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L170).  See also https://xrpl.org/escrowcancel.html

### Context of Change

### Type of Change

## Before / After
`XRPEscrowCancel` class and conversion unit tests now exist.

## Test Plan


XpringKit/XRP/Model/XRPEscrowCancel+Org_Xrpl_Rpc_V1_EscrowCancel.swift
```diff
@@ -0,0 +1,29 @@
...
+  /// src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L170)
+  ///
+  /// - Parameters:
+  ///     - escrowCancel: an Org_Xrpl_Rpc_V1_EscrowCancel (protobuf object) whose field values will be used to
+  ///             construct an XRPEscrowCancel
+  /// - Returns: an XRPEscrowCancel with its fields set via the analogous protobuf fields.
+  init?(escrowCancel: Org_Xrpl_Rpc_V1_EscrowCancel, xrplNetwork: XRPLNetwork) {
+    guard let ownerXAddress = Utils.encode(
+      classicAddress: escrowCancel.owner.value.address,
+      isTest: xrplNetwork != XRPLNetwork.main
```
tedkalaw, line 19: worth adding a helper like i did? i kno u mentioned it in my pr but was curious if you're gonna do that b4/after the wrappers are done

keefertaylor: +1 on this being a good idea. 

You could just add it as an extension on the XRPLNetwork enum. 

```swift
extension XRPLNetwork {
  public var isTest: Bool {
     return self != .Main
  }
}
```

amiecorso: Yep - going to actually do this all at once when all the wrappers are in

PR #282, opened 2020-08-24
(13/17) Creates `XRPPaymentChannelCreate` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPPaymentChannelCreate` - a native Swift struct to wrap the [`PaymentChannelCreate` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L272).  See also https://xrpl.org/paymentchannelcreate.html

### Context of Change

### Type of Change

## Before / After
`XRPPaymentChannelCreate` class and conversion unit tests now exist.

## Test Plan


Tests/XRP/Helpers/TransactionTypeTestProtos/Org_Xrpl_Rpc_V1_PaymentChannelCreate+Test.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+    $0.settleDelay = Org_Xrpl_Rpc_V1_SettleDelay.with {
+      $0.value = .testSettleDelay
+    }
+    $0.publicKey = Org_Xrpl_Rpc_V1_PublicKey.with {
+      $0.value = .testPublicKey
+    }
+  }
+
+  public static let testPaymentChannelCreateMissingDest = Org_Xrpl_Rpc_V1_PaymentChannelCreate.with {
```
tedkalaw, line 44: i know keefer doesn't like abbreviations - `Destination`?

amiecorso: getting yelled at for a variable name that's too long... since it's test code I bet Keefer will let it slide

keefertaylor: oof. That's fine with me. We could also disable this linter rule globally (if it doesn't provide value) or disable it for tests (if it isn't worth enforcing) or just for this line (if we want a one off).

PR #283, opened 2020-08-25
(14/17) Creates `XRPPaymentChannelFund` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPPaymentChannelFund` - a native Swift struct to wrap the [`PaymentChannelFund` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L288).  See also https://xrpl.org/paymentchannelfund.html

### Context of Change

### Type of Change

## Before / After
`XRPPaymentChannelFund` class and conversion unit tests now exist.

## Test Plan


XpringKit/XRP/Model/XRPPaymentChannelFund+Org_Xrpl_Rpc_V1_PaymentChannelFund.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    if paymentChannelFund.hasAmount {
+      if let amount = XRPCurrencyAmount(currencyAmount: paymentChannelFund.amount.value) {
+        self.amount = amount
+      } else {
+        return nil
+      }
+    } else {
+      return nil
+    }
```
tedkalaw, line 31: not sure what preferred style is in `XpringKit` but if you invert the top two logical statements and return early, you'll lose a level of indentation ie

```
if !paymentChannelFund.hasChannel {
  return nil
}
self.channel = ....
```

i have strong pref for returning early but up to ya'll

keefertaylor: Same. 

You can even us a `guard` here which will programmatically enforce you to return early. 

```
guard paymentChannelFund.hasChannel else {
}
```

amiecorso: great - i'm going to update this logic throughout the entire set of classes in this PR: https://github.com/xpring-eng/XpringKit/pull/288

PR #284, opened 2020-08-25
(15/17) Creates `XRPSetRegularKey` Protobuf Wrapper Class

amiecorso:
This PR implements the class `XRPSetRegularKey` - a native Swift class to wrap the [`SetRegularKey` protocol buffer](https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L298).  See also https://xrpl.org/setregularkey.html

### Context of Change

### Type of Change

## Before / After
`XRPSetRegularKey` class and conversion unit tests now exist.

## Test Plan


XpringKit.xcodeproj/project.pbxproj
```diff
@@ -1551,8 +1562,6 @@
 			isa = PBXProject;
 			attributes = {
 				LastUpgradeCheck = 1020;
-				TargetAttributes = {
-				};
```
tedkalaw, line 1555: out of curiosity, was this just unused from a previous something?

keefertaylor: This file is autogenerated when you run `xcodegen`. 

It's likely that this changes based on what exact version of XCode you have installed. The file which generates this file is `project.yml` and that's the canonical definition.

